PR #8, opened 2017-08-26
Fixed missing alignment on MinGW + GCC

smx-smx:
MinGW defines _WIN32, so alignment is missing when using GCC.
This patch switches to _MSC_VER, so that GCC is handled properly even on MinGW/Windows


recp:
@smx-smx thanks for your contribution!

Merged.

PR #14, opened 2017-12-30
feature: extract view frustum planes and corners

recp:
I think it would be nice to support extracting planes in **cglm** as built-in feature.

It uses **Gil Gribb** and **Klaus Hartmann** paper: http://gamedevs.org/uploads/fast-extraction-viewing-frustum-planes-from-world-view-projection-matrix.pdf

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -52,3 +52,9 @@ LICENSE:
 
   Computing Euler angles from a rotation matrix (euler.pdf)
   Gregory G. Slabaugh
+
+4. Extracting Planes
+Fast Extraction of Viewing Frustum Planes from the World-View-Projection Matrix
+Authors:
+  Gil Gribb (ggribb@ravensoft.com)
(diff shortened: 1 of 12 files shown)


recp:
`glm_extract_planes` 
`glm_extract_corners` 

vs 

`glm_frustum_planes`
`glm_frustum_corners` 

? 

`glm_extract_planes` which planes?  The second one seems more meaningful, also will be namespace for new frustum-specific features 👍

winduptoy:
+1 for the namespacing of `glm_frustum_xxx`. I think namespace hierarchy like that (discoverability, autocomplete), is far more important than matching the syntax of proper English.

recp:
I used `vec4` for storing corners, I hope this will not be a problem for someone others. To converting it back to `vec3` requires a little bit work. Also if we have to compute corners and center on every camera moving then maybe we must choice performance over memory (extra 8 byte)

PR #16, opened 2018-01-04
convenient functions for perspective projection matrix

recp:
- `glm_persp_fovy` extracts fovy
- `glm_persp_aspect` extracts aspect ratio
- `glm_persp_sizes ` extracts [nearWidth, nearHeight, farWidth, farHeight]
- extracting near, far, left, light, top an bottom functions are optimized (not too much)

### DIFF
--- a/include/cglm/cam.h
+++ b/include/cglm/cam.h
@@ -427,6 +427,31 @@ glm_persp_decomp_near(mat4 proj, float * __restrict nearVal) {
 }
 
 /*!
+ * @brief returns field of view angle along the Y-axis (in radians)
+ *
+ * if you need to degrees, use glm_deg to convert it or use this:
+ * fovy_deg = glm_deg(glm_persp_fovy(projMatrix))
+ *
+ * @param[in] proj perspective projection matrix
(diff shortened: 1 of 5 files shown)


include/cglm/cam.h
```diff
@@ -323,12 +323,28 @@ glm_persp_decomp(mat4 proj,
...
+  m22 = proj[2][2];
+  m32 = proj[3][2];
+
+  f = m32 / (m22 - 1);
+  n = m32 / (m22 + 1);
+
+  n_m11 = n / m11;
+  n_m00 = n / m00;
+
+  *nearVal = f;
```
recp: todo: typo I must fix it, it must be n

recp:
There are some floating point error problems here, we should not build something on decomposing projection matrix for production (for now) especially for near/far values. 

Check this: https://stackoverflow.com/questions/48135014/unpredictable-floating-point-errors-from-matrix-decomposition

In the future I'll try to fix the floating point errors. Actually it works for small far distances (100 is common I guess) but it must work consistently

PR #18, opened 2018-01-10
new wrappers for lookat

recp:
New vector function:
```C
void glm_vec_ortho(vec3 v, vec3 dest);
```

New camera functions:
```C
void glm_look(vec3 eye, vec3 dir, vec3 up, mat4 dest);
void glm_look_any(vec3 eye, vec3 dir, mat4 dest);
```

Sometimes we have only direction vector to create view matrix. In this case we do this:

```C
vec3 target;
glm_vec_add(eye, dir, target);
glm_lookat(eye, target, up, dest);
```

because we need to target/center to construct view matrix with `lookat`, if we have only direction then target/center we want to create will be fake... 

So let cglm do this for us. `glm_look ` is just wrapper for original `lookat` which accepts `direction` instead of `center/target`. It creates center/target for us.

`glm_vec_ortho` creates possible orthogonal/perpendicular vector for given vector. `look_any` uses this vector as UP vector as orthogonal to direction vector. So if it is not matter what UP vector is, then this makes things easier. 

R

### DIFF
--- a/include/cglm/call/cam.h
+++ b/include/cglm/call/cam.h
@@ -43,10 +43,15 @@ glmc_perspective(float fovy,
 
 CGLM_EXPORT
 void
-glmc_lookat(vec3 eye,
-            vec3 center,
-            vec3 up,
-            mat4 dest);
+glmc_lookat(vec3 eye, vec3 center, vec3 up, mat4 dest);
+
+CGLM_EXPORT
+void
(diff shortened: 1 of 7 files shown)


recp:
Any good name for these functions are welcome!

recp:
Alternative names for `look_any`

1. `glm_look_anyup`
2. `glm_look_noup`

`glm_look_anyup` seems more attractive because unlike `glm_look_any` it implies what the function do

PR #19, opened 2018-01-12
move frustum related stuff to separate frustum header

recp:
New:
* func for get bounding box of frustum

Improvements:
* create helpers macro which defines corner index
* add missing source to make file
* add more desc to glm_frustum_corners


recp:
Here how I created shadow matrix for directional light in upcoming [graphics / render engine](http://github.com/recp/gk) which also written with C

Part 1: Update camera properties if camera moved/transformed
```C
mat4 invViewProj;

glm_mat4_inv(cam->viewProj, invViewProj);

glm_frustum_planes(cam->viewProj,        cam->frustum.planes);
glm_frustum_corners(invViewProj,         cam->frustum.corners);
glm_frustum_center(cam->frustum.corners, cam->frustum.center);
```

Part 2: Get view and projection matrices for directional light for shadowmap
```C
mat4 view, proj;

glm_look_anyup(cam->frustum.center, light->dir, view); /* light POV  (arbitrary UP axis) */
glm_frustum_box(cam->frustum.corners, view, box); /* Bounding box */

glm_ortho(box[0][0], box[1][0], 
          box[0][1], box[1][1], 
          box[0][2], box[1][2], proj);

glm_mat4_mul(proj, view, viewProj);
```

Splitting frustum requires some extra stuff, it can be found in gk (currently in local) in the future

Some ( maybe all :) ) math may wrong or needs to be improved, I will fix/improve it by time.

recp:
Now we have `glm_frustum_corners_at` function which will help us to split frustum. It takes `frustum corners` which created previously, `split distance (from near)` and `far` then gives planes' corners. 

Example usage:
```C
glm_frustum_corners_at(frustum->corners,
                       distance,
                       farDistance,
                       &subFrustum.corners[4]);
```

To get a sub frustum we need another plane but we already have which is near or far or previously created plane by this function ;)

Now my only concern is that all corners must be `vec4`, since it is comfortable for cglm but user must convert `vec3` corners to `vec4` before use or after use. In the future this may change

Another possible change will be `glm_frustum_corners` because it uses clip space coords but it may change to use geometric approach

Possible name changes:
`glm_frustum_corners` -> `glm_frustum_corners_cs` (clip space)
`glm_frustum_corners` (geometrix approach) (default)

The last thing is that floating point errors may occur especially if far is too far. It is in my TODOs but currently I don't know what to do about that without using double. There must be some math to fix it, in the future I will try to fix that by learning standards about floats. Now it is postponed.

PR #20, opened 2018-01-14
add new headers to makefile.am

sdstrowes:
Clean build didn't include these headers in the install location. Adding them into makefile.am so they're automatically put in place.


recp:
Hi @sdstrowes, thanks for your contribution! 

Could you do same for call version too (cglm_call_HEADERS)?

recp:
@sdstrowes thanks! Merged.

PR #23, opened 2018-01-18
bounding box

recp:
New functions:
`glm_aabb_transform` - transforms aabb with matrix
`glm_aabb_merge` - merges two axis aligned bounding boxes, creates combined one. 
`glm_aabb_crop` - crop a bbox with another one
`glm_aabb_crop_until`

especially `glm_aabb_crop` is fantastic! e.g. if you have bbox for view frustum and your scene is smaller than view frustum's bbox then you can shrink frustum bbox to scene bbox. If scene bbox is larger than (one edge or all edges) then it picks smaller one. In the end view frustum bbox will be max.

to get scene bbox (pseudo code):
```
sceneBbox = objects[0].bbox
for (i = 1; i < objects->count; i++)
  glm_aabb_merge(sceneBbox, objects[i].bbox, sceneBbox)
```

if you do same for frustum after culled, then you get scene bbox in frustum, scene bbox may larger or may smaller so `glm_aabb_crop` will help to crop it:
```
sceneBbox = frustumObjects[0].bbox
for (i = 1; i < frustumObjects->count; i++)
  glm_aabb_merge(sceneBbox, frustumObjects[i].bbox, sceneBbox)

finalBbox
glm_aabb_crop(frustumBbox, sceneInFrustumBbox, finalBbox)
```

`glm_aabb_transform` is also optimized, I saw it at: http://dev.theomader.com/transform-bounding-boxes/

and implemented in cglm. 

these functions may need more tests and more tests with different cases...

TODOs for future: 
- `glm_aabb_intersects` - check if two bbox intersects or not (boolean)
- `glm_aabb_contains` - check if bbox contains/encloses other bbox or not (boolean)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -58,3 +58,8 @@ Fast Extraction of Viewing Frustum Planes from the World-View-Projection Matrix
 Authors:
   Gil Gribb (ggribb@ravensoft.com)
   Klaus Hartmann (k_hartmann@osnabrueck.netsurf.de)
+
+5. Transform AABB
+Transform Axis Aligned Bounding Boxes:
+http://dev.theomader.com/transform-bounding-boxes/
+https://github.com/erich666/GraphicsGems/blob/master/gems/TransBox.c
--- a/LICENSE
+++ b/LICENSE
@@ -63,3 +63,7 @@ Authors:
 Transform Axis Aligned Bounding Boxes:
 http://dev.theomader.com/transform-bounding-boxes/
 https://github.com/erich666/GraphicsGems/blob/master/gems/TransBox.c
+
+6. Cull frustum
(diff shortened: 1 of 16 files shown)


recp:
now we have `glm_aabb_frustum` (source: http://www.txutxi.com/?p=584)
function which will help to cull  frustum. it was `glm_aabb_in_frustum` but I like short names. 

```C
if (glm_aabb_frustum(box, camera->planes)) {
  /* object intersects with frustum, gather */
}

/* or */

if (!glm_aabb_frustum(box, camera->planes)) {
  /* outside! cull it */
}
```

Actually I thought that maybe we can use `glm_abb_` instead of `glm_aabb_` but the second one seems more readable

Issue #30, opened 2018-02-27
Euler angles proposal


recp:
Ken Shoemake's algorithm seems most common one, it also handles 24 different permutations of rotation order. Maybe instead of providing separate functions for euler angles we could adopt that algorithm / codes. 

Related implementation (Graphics Gems IV): https://github.com/erich666/GraphicsGems/tree/master/gemsiv/euler_angle
https://github.com/erich666/GraphicsGems/blob/master/gemsiv/euler_angle/EulerAngles.c

It also provides euler to quaternion which is in our TODO list. I hope there is no any LICENSE conflict

PR #43, opened 2018-04-07
quaternion improvements and new features

recp:
New functions: 
- [x] `glm_quat_mat4t(versor q, mat4 dest)` - convert quat to mat4 (transposed)
- [x] `glm_quat_mat3(versor q, mat3 dest)` - convert quat to mat3 
- [x] `glm_quat_mat3t(versor q, mat3 dest)` - convert quat to mat3  (transposed)
- [x] `glm_mat4_quat(mat4 m, versor dest)` - convert mat4 to quat
- [x] `glm_mat3_quat(mat3 m, versor dest)` - convert mat3 to quat
- [x] `glm_quat_conjugate(versor q, versor dest)`
- [x] `glm_quat_inv(versor q, versor dest)` - inverse of quaternion
- [x] change **WXYZ** order to **XYZW**. (In Progress)
- [x] change `quat_mulv`to `quat_mul`(because we multiply quat with quat, not vector), we may use `mulv` for multiply quat with vector (In Progress)
- [x] SSE/SSE2 version for quat_mul (maybe)
- [x] `glm_quat_axis(versor q, vec3 dest)` - get axis from quat
- [x] `float glm_quat_real(versor q)` - returns w
- [x] `float glm_quat_angle(versor q)` - get angle from quat
- [x] `glm_quat_imag(versor q, vec3 dest)`
- [x] `glm_quat_imagn(versor q, vec3 dest)` - normalized version of `_imag
- [x] `float glm_quat_imaglen(versor q)`
- [x] `glm_quat_add(versor p, versor q, versor dest) `
- [x] `glm_quat_sub(versor p, versor q, versor dest) `
- [x] rotate vector with quat (`glm_quat_roatev(versor q, vec3 v, vec3 dest)` maybe)
- [x] rotate matrix with quat (`glm_quat_roate(versor q, mat4 transform, mat4 dest)` maybe)
- [x] update docs (In Progress)
- [x] tests (In Progress)
- [x] `glm_quat_look(vec3 eye, versor ori, mat4 dest)` - create view matrix using quaternion
- [x] `glm_quat_for(vec3 dir, vec3 fwd, vec3 up, versor dest)` - creates quaternion for look rotation (source to dest point) 

New Helpers:
- [x] `glm_vec4_sqrt(vec4 v, vec4 dest)` 
- [x] `glm_vec_sqrt(vec3 v, vec3 dest)`
- [x] `glm_vec4_sign(vec4 v, vec4 dest)` 
- [x] `glm_vec_sign(vec3 v, vec3 dest)`
- [x] SSE2 implementation for vec4_dot and vec4_norm2
- [x] `glm_vec_lerp(vec3 from, vec3 to, float t, vec3 dest)`
- [x] `glm_vec4_lerp(vec4 from, vec4 to, float t, vec4 dest)`
- [x] `float glm_lerp(float from, float to, float t)`
**[DROPPED]** - [x] `glm_mat4_mulq(mat4 m, versor q, mat4 dest)` - multiply mat4 with quat; now we have `glm_quat_roate` for this purpose
**[DROPPED]** - [x] `glm_mat4_quat_sse2(mat4 m, versor dest)` - convert mat4 to quat (SSE2 version)
**[DROPPED]** - [x] `glm_quat_slerp_sse2(versor from, versor to, float t, versor dest) ` - it seems it is not produce correct results, also `glm_quat_slerp` version is improved and using `vec4`s no need to write manual SSE codes because `vec4` functions are optimized with SIMD functions
**[POSTPONED]** - [ ] `float glm_quat_euler(versor q, vec3 eulXYZ)` - convert quat to euler angles
**[POSTPONED]** - [ ] `float glm_euler_quat(vec3 eulXYZ, versor dest)` - convert euler angles to quat
---

**CRITICAL CHANGES:**
- normalize quaternion (not input itself) in `glm_quat_mat4` and `glm_quat_mat3` before convert to matrix (if this is not good idea, please let me know)
- change **WXYZ** order to **XYZW**
- change `glm_quat_mulv` to `glm_quat_mul` and use `glm_quat_mulv`for multiply quat with vector

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,4 @@ cglm_test_iosTests/*
 docs/build/*
 win/cglm_test_*
 * copy.*
+*.o
--- a/CREDITS
+++ b/CREDITS
@@ -46,3 +46,7 @@ http://old.cescg.org/CESCG-2002/DSykoraJJelinek/
 
 7. Quaternions
 Initial mat4_quat is borrowed from Apple's simd library
+
+
+8. Vector Rotation using Quaternion
+https://gamedev.stackexchange.com/questions/28395/rotating-vector3-by-a-quaternion
--- a/CREDITS
+++ b/CREDITS
@@ -43,3 +43,6 @@ https://github.com/erich666/GraphicsGems/blob/master/gems/TransBox.c
 6. Cull frustum
 http://www.txutxi.com/?p=584
 http://old.cescg.org/CESCG-2002/DSykoraJJelinek/
+
+7. Quaternions
+Initial mat4_quat is borrowed from Apple's simd library
--- a/configure.ac
+++ b/configure.ac
@@ -7,7 +7,7 @@
 #*****************************************************************************
 
 AC_PREREQ([2.69])
-AC_INIT([cglm], [0.3.6], [info@recp.me])
+AC_INIT([cglm], [0.4.0], [info@recp.me])
 AM_INIT_AUTOMAKE([-Wall -Werror foreign subdir-objects])
 
 AC_CONFIG_MACRO_DIR([m4])
--- a/docs/source/quat.rst
+++ b/docs/source/quat.rst
@@ -5,17 +5,16 @@ quaternions
 
 Header: cglm/quat.h
 
- **Important:** *cglm* stores quaternion as [w, x, y, z] in memory, don't
- forget that when changing quaternion items manually. For instance *quat[3]*
- is *quat.z* and *quat[0*] is *quat.w*. This may change in the future if *cglm*
- will got enough request to do that. Probably it will not be changed in near
- future
+ **Important:** *cglm* stores quaternion as **[x, y, z, w]** in memory
+ since **v0.4.0** it was **[w, x, y, z]**
+ before v0.4.0 ( **v0.3.5 and earlier** ). w is real part.
 
-There are some TODOs for quaternions check TODO list to see them.
+What you can do with quaternions with existing functions is (Some of them):
 
-Also **versor** is identity quaternion so the type may change to **vec4** or
-something else. This will not affect existing functions for your engine because
-*versor* is alias of *vec4*
+- You can rotate transform matrix using quaterion
+- You can rotate vector using quaterion
+- You can create view matrix using quaterion
+- You can create a lookrotation (from source point to dest)
 
 Table of contents (click to go):
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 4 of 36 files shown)


recp:
I implemented in "Alternative Method" mentioned in: http://www.euclideanspace.com/maths/geometry/rotations/conversions/matrixToQuaternion/index.htm

```C
quaternion.w = sqrt( max( 0, 1 + m00 + m11 + m22 ) ) / 2;
quaternion.x = sqrt( max( 0, 1 + m00 - m11 - m22 ) ) / 2;
quaternion.y = sqrt( max( 0, 1 - m00 + m11 - m22 ) ) / 2;
quaternion.z = sqrt( max( 0, 1 - m00 - m11 + m22 ) ) / 2;
Q.x = _copysign( Q.x, m21 - m12 )
Q.y = _copysign( Q.y, m02 - m20 )
Q.z = _copysign( Q.z, m10 - m01 )
```

It requires extra `sqrt` functions but for since it requires 4 `sqrt` I used SIMD instruction here. So without function call and with SINGLE instruction we calculate all `sqrt` which is good. This is also true for `sign` function. **cglm**'s sign functions return -1, 0 and +1 which makes the algorithm work. 

I didn't compare performances between other method and this "alternative" method but since it is possible to write SIMD version I used this one. `glm_mat4_quat_sse2` is SIMD version which will be called if SSE2 is supported. No need to call this manually

In the future we may change this if we find better approach...

Also I found this one: https://d3cw3dd2w32x2b.cloudfront.net/wp-content/uploads/2015/01/matrix-to-quat.pdf which seems interesting but I didn't worke for me, I don't know.

recp:
One another thing is that, `glm_quat_mat3` and `glm_quat_mat4` functions didn't normalize the input quaternion, now they normalize the input quat. Because it must be unit quaternion and most user will not aware of this and will get different rotation matrices. 

**glm** don't normalize input quaternion if I read it correctly and I don't know why: https://github.com/g-truc/glm/blob/8390a77b3a278b15259e5ca6e67f7e41badc457b/glm/gtc/quaternion.inl#L619

We should normalize it, and maybe we should provide alternative function which accepts only unit vector.

recp:
Should we change the order of quaternion members from `wxyz` to `xyzw`? Any feedback?

recp:
**Update:** I updated matrix to quat code, I was said that I used alternative method but it didn't work well in tests. So I switched to other one, I borrowed mat2quat from Apple's **simd** library which works well and passes all tests.

shakesoda:
switching quat members to xyzw would be consistent with some other math libraries I've used, and my own internal ones. 👍

recp:
@shakesoda thanks for your feedback[s] 🤗, I thought the same, most libraries use xyzw and making cglm that way would make easier to copy quaternion from one library to another.

Especially new-comers may assume that **cglm** uses **xyzw** (it is wxyz for now) but if they construct quaternion manually they will create wrong one and get wrong rotation matrix... **wxyz** perfectly matches with math expression (w + xi + yj + zk) but **I'll change/update it to **xyzw** in this PR** 👍 because in GLSL-like lang/apis quat.w is last element...

One another thing I want to get feedback is that I want to change `versor` type to `quat`, but unlike `vec3`, `vec4`, `mat4`... the **quat** type not seems like a type, it seems like a variable. I don't want conflicts between user variables and **cglm** types (and system types/macros). 

Alternative names:
`quat`
`quat32`
`quat_t`

or we just continue to use `versor` or just `vec4` I'm not sure which one is better. (`versor` is a type alias of `vec4`)

PS: I dropped "alternative method" but I created a gist (https://gist.github.com/recp/97d87c8b9c3fe8ac432a865dff3c783d) for it, in the future we may bring it back if we can fix it's numerical errors (probably we will not)

recp:
Here the new lookat function:

`void glm_quat_look(vec3 eye, versor ori, mat4 dest)`

it is similar to `glm_lookat`and `glm_look` but unlike them this one creates view matrix using quaternion. Quaternion specifies UP and RIGHT vectors. Also it generates less instructions than `glm_lookat`and `glm_look`

Also I postponed quaternion -> euler because we **may** change euler API later, check this issue: https://github.com/recp/cglm/issues/30

recp:
Now we also have another option to rotate vectors: **Quaternions**! 

New option:
- `glm_quat_rotatev(versor q, vec3 v, vec3 dest)` - rotate vector using quat

Previous options:
- `glm_vec_rotate(vec3 v, float angle, vec3 axis)` - rotate vector using angle and axis
- `glm_vec_rotate_m4(mat4 m, vec3 v, vec3 dest)` - rotate vector using matrix
- manual matrix-vector  multiplication

Also, added some tests for quaternions everthing looks good except I didn't add tests for slerp and lerp. 

- **[bugfix]** While comparing `glm_quat_rotatev` and `glm_vec_rotate` results I have realized that we should normalize axis in `glm_quatv` and `glm_vec_rotate`. Now they are normalized and we get same result in both `glm_vec_rotate` and `glm_quat_rotatev`

recp:
`glm_mat4_mulq` is replaced with `glm_quat_rotate(mat4 m, versor q, mat4 dest)` because `glm_quat_rotate` seems better name. It rotates existing transform using quaternion

@winduptoy IIRC we have discussed quat order before, I hope you caught **WXYZ** to **XYZW** changes and I hope you are ok with this too

I added some tests (and lots of APIs) and everthing looks good, we have better quaternion APIs now. 

I'm going to merge this PR after updated docs and change version from v0.3.5 to v0.4.0

PR #49, opened 2018-04-30
replace _WIN32 with _MSC_VER

Yatima1460:
Compiles fine under MinGW 64 on Windows too


recp:
Thank you very much for your contributions! I think checking **_MSC_VER** is better than WIN32. since it works on mingw, everthings look good!

PR #59, opened 2018-07-18
Fix alignment issue when using AVX

jonathanplatzer:
When AVX is used, the default alignment for `mat3` and `mat4` is still 16 Bytes, which produces a segfault when used, because the `_mm256_load_ps` and `_mm256_store_ps` operations require 32 Byte aligned memory.

This commit fixes this issue by adding a conditional to `types.h` that checks if AVX is used, and if so, aligns `mat3` and `mat4` by 32 Bytes.

It also removes the `CGLM_ALIGN(16)` directive from `glm_mat4_identity` in `mat4.h`, where I encountered this bug, so the default alignment from `types.h` will be used.

### DIFF
--- a/include/cglm/affine-mat.h
+++ b/include/cglm/affine-mat.h
@@ -151,7 +151,7 @@ glm_inv_tr(mat4 mat) {
 #if defined( __SSE__ ) || defined( __SSE2__ )
   glm_inv_tr_sse2(mat);
 #else
-  CGLM_ALIGN(16) mat3 r;
+  CGLM_ALIGN_MAT mat3 r;
   CGLM_ALIGN(16) vec3 t;
 
   /* rotate */
--- a/include/cglm/affine.h
+++ b/include/cglm/affine.h
@@ -244,7 +244,7 @@ glm_scale_uni(mat4 m, float s) {
 CGLM_INLINE
 void
 glm_rotate_x(mat4 m, float angle, mat4 dest) {
-  CGLM_ALIGN(16) mat4 t = GLM_MAT4_IDENTITY_INIT;
+  CGLM_ALIGN_MAT mat4 t = GLM_MAT4_IDENTITY_INIT;
   float c, s;
 
   c = cosf(angle);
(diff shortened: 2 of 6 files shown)


jonathanplatzer:
> Maybe instead of dropping them we could use 32-byte alignment on stack variables if AVX enable?

Yes, that came to my mind a little bit afterwards.

I thought about doing something like this in `glm_mat4_identity`
```c
#if __AVX__
  CGLM_ALIGN(32) mat4 t = GLM_MAT4_IDENTITY_INIT;
#else
  CGLM_ALIGN(16) mat4 t = GLM_MAT4_IDENTITY_INIT;
#endif
```
I think this would be the only place in the code where this needed, since I haven't seen any other AVX intrinsics where a new variable is used.

recp:
What about using macros like this:

```C
/* types or common header */

#ifdef __AVX__
#define CGLM_ALIGN_VEC4 CGLM_ALIGN(32)
#else
#define CGLM_ALIGN_VEC4 CGLM_ALIGN(16)
#endif

/* inside function */
CGLM_ALIGN_VEC4 mat4 t = GLM_MAT4_IDENTITY_INIT;
```

I'm not sure, maybe we should drop all alignment attribs on stack variables. I'm confused a bit, because if we use 32-byte alignment on vec4 if AVX enabled; some AVX instructions can fill AVX  register using SSE registers, in this case maybe 16-byte alignment would be enough. 

Maybe we should only use 32-byte alignment on mat4/mat3. In this case changing above definitions to `CGLM_ALIGN_MAT` seems good to me. 

If we want to keep alignment on stack variables, there are a few CGLM_ALIGN(16) to be changed. Some of them are in affine.h some of them are in mat3.h

jonathanplatzer:
`CGLM_ALGIN_MAT` only on mat4 and mat3 is a good idea.
I can add another commit with it to the PR.

recp:
Thanks 🤗

recp:
@jonathanplatzer merged, thanks for your contribution!

PR #61, opened 2018-09-19
Replace non-standard M_PI* constants

hartenfels:
The constants `M_PI`, `M_PI_2` and `M_PI_4` are not in the C standard. If you, for example, run GCC in strict C99 mode, including CGLM headers will fail to compile.

A minimal example to demonstrate, let's call it `mpi.c`:

```c
#include <cglm/cglm.h>
int main(void) {       
    return 0;
}
```

Trying to compile this with `gcc -std=c99 mpi.c` will barf saying that `M_PI` isn't defined.

```
$ gcc -std=c99 mpi.c
In file included from /usr/local/include/cglm/common.h:57:0,
                 from /usr/local/include/cglm/cglm.h:11,
                 from mpi.c:1:
/usr/local/include/cglm/util.h: In function ‘glm_rad’:
/usr/local/include/cglm/util.h:61:16: error: ‘M_PI’ undeclared (first use in this function); did you mean ‘__P’?
   return deg * CGLM_PI / 180.0f;

(And like a dozen more errors like this.)
```

This pull request simply replaces those constants with their literal numeric values. I retained the cast from double to float to make sure there's really no change in behavior.

For comparison, the definitions in [musl libc](https://github.com/cloudius-systems/musl/blob/00733dd1cf791d13ff6155509cf139a5f7b2eecb/include/math.h#L363-L365) and [glibc](https://github.com/lattera/glibc/blob/master/math/math.h#L1065-L1067) are identical.

### DIFF
--- a/include/cglm/types.h
+++ b/include/cglm/types.h
@@ -48,8 +48,17 @@ typedef CGLM_ALIGN_IF(16) vec4  mat4[4];
 
 typedef vec4                    versor;
 
-#define CGLM_PI    ((float)3.14159265358979323846264338327950288)
-#define CGLM_PI_2  ((float)1.57079632679489661923132169163975144)
-#define CGLM_PI_4  ((float)0.785398163397448309615660845819875721)
+#define GLM_PI     3.14159265358979323846264338327950288
+#define GLM_PI_2   1.57079632679489661923132169163975144
+#define GLM_PI_4   0.785398163397448309615660845819875721
+
+#define GLM_PIf    ((float)GLM_PI)
+#define GLM_PI_2f  ((float)GLM_PI_2)
+#define GLM_PI_4f  ((float)GLM_PI_4)
+
+/* DEPRECATED! use GLM_PI and friends */
+#define CGLM_PI    GLM_PIf
+#define CGLM_PI_2  GLM_PI_2f
+#define CGLM_PI_4  GLM_PI_4f
 
 #endif /* cglm_types_h */
--- a/include/cglm/types.h
+++ b/include/cglm/types.h
@@ -48,8 +48,8 @@ typedef CGLM_ALIGN_IF(16) vec4  mat4[4];
 
 typedef vec4                    versor;
 
-#define CGLM_PI    ((float)3.14159265358979323846)
-#define CGLM_PI_2  ((float)1.57079632679489661923)
-#define CGLM_PI_4  ((float)0.78539816339744830962)
+#define CGLM_PI    ((float)3.14159265358979323846264338327950288)
+#define CGLM_PI_2  ((float)1.57079632679489661923132169163975144)
+#define CGLM_PI_4  ((float)0.785398163397448309615660845819875721)
 
 #endif /* cglm_types_h */
--- a/include/cglm/types.h
+++ b/include/cglm/types.h
@@ -45,7 +45,6 @@ typedef                   vec3  mat3[3];
 typedef CGLM_ALIGN_IF(16) vec4  mat4[4];
 #endif
 
-
 typedef vec4                    versor;
 
 #define GLM_E         2.71828182845904523536028747135266250   /* e              */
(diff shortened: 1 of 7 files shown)


recp:
@hartenfels thanks for your contribution!


What about checking existence of M_PI e.g. `#ifndef M_PI` and others before defining it manually? Since `float` does not have very high precision, the current change also looks good!

> I retained the cast from double to float to make sure there's really no change in behavior.

casting to float was used because original macro is double (or long double). The main goal here was suppressing warnings

We could define it as float literal like this:
```C
#define CGLM_PI    3.14159265358979323846f
```

My macOS defined PI numbers like this in `math.h`:

```C
/*  Even though these might be more useful as long doubles, POSIX requires
    that they be double-precision literals.                                   */
#define M_PI        3.14159265358979323846264338327950288   /* pi             */
#define M_PI_2      1.57079632679489661923132169163975144   /* pi/2           */
#define M_PI_4      0.785398163397448309615660845819875721  /* pi/4           */
```

Maybe defining them double may help to get better precision for some CPUs (maybe not today but in the future)? 

Another idea (maybe for the future):

Since all **cglm** macros (except configs) uses GLM_ prefix we can convert this name into GLM_PI, GLM_PI_2... by deprecating CGLM_PI... versions.

hartenfels:
Checking with `#ifdef` was actually the first thing I did, but then it just looked so ugly that I went for just using the numbers directly. Pi *probably* isn't gonna change anytime soon, so just putting the number in there should do the trick.

Using floating point literals would *probably* be fine, but I'm not that deep into the C standard that I can say for certain that it's always the exact same result. But I can change them over to float literals to make it look nicer if you want.

Renaming it to `GLM_PI` and friends seems pretty sensible to me for consistency's sake, as long as the old definitions stick around for a while.

recp:
My suggestion:

```C
#define GLM_PI     3.14159265358979323846264338327950288
#define GLM_PI_2   1.57079632679489661923132169163975144
#define GLM_PI_4   0.785398163397448309615660845819875721

#define GLMF_PI    ((float)GLM_PI)
#define GLMF_PI_2  ((float)GLM_PI_2)
#define GLMF_PI_4  ((float)GLM_PI_4)

/* DEPRECATED! use GLM_PI and friends */
#define CGLM_PI    GLMF_PI
#define CGLM_PI_2  GLMF_PI_2
#define CGLM_PI_4  GLMF_PI_4
```

**F** suffix/postfix can be changed or moved elsewhere

In this way we define PI both for double and float

include/cglm/types.h
```diff
@@ -48,8 +48,17 @@ typedef CGLM_ALIGN_IF(16) vec4  mat4[4];
 
 typedef vec4                    versor;
 
-#define CGLM_PI    ((float)M_PI)
-#define CGLM_PI_2  ((float)M_PI_2)
-#define CGLM_PI_4  ((float)M_PI_4)
+#define GLM_PI     3.14159265358979323846264338327950288
+#define GLM_PI_2   1.57079632679489661923132169163975144
+#define GLM_PI_4   0.785398163397448309615660845819875721
```
recp: Since we defined math constants **manually** and since it might not be defined in anywhere else maybe defining other numbers would also help developers who use them. 

I think defining other numbers here is not bad idea. Other definitions:

```C
#define GLM_E         2.71828182845904523536028747135266250   /* e              */
#define GLM_LOG2E     1.44269504088896340735992468100189214   /* log2(e)        */
#define GLM_LOG10E    0.434294481903251827651128918916605082  /* log10(e)       */
#define GLM_LN2       0.693147180559945309417232121458176568  /* loge(2)        */
#define GLM_LN10      2.30258509299404568401799145468436421   /* loge(10)       */
#define GLM_PI        3.14159265358979323846264338327950288   /* pi             */
#define GLM_PI_2      1.57079632679489661923132169163975144   /* pi/2           */
#define GLM_PI_4      0.785398163397448309615660845819875721  /* pi/4           */
#define GLM_1_PI      0.318309886183790671537767526745028724  /* 1/pi           */
#define GLM_2_PI      0.636619772367581343075535053490057448  /* 2/pi           */
#define GLM_2_SQRTPI  1.12837916709551257389615890312154517   /* 2/sqrt(pi)     */
#define GLM_SQRT2     1.41421356237309504880168872420969808   /* sqrt(2)        */
#define GLM_SQRT1_2   0.707106781186547524400844362104849039  /* 1/sqrt(2)      */
```

I borrowed this in `math.h` header in macOS and added `GL` prefix to them. We can use them with their comments.

If using all is redundant for maintenance then we could pick related numbers like 1/PI, 2/PI... What do you think? 

It is not required in this PR, I can add this later if you want.

hartenfels: Whoops, missed those somehow. Fixed that now.

I guess I'll just add those numbers as well, while we're at it. None of those `M_` constants is in standard C, so it makes a lot of sense for a math library like this to provide them.

hartenfels:
Uh, looks like the comment I'd posted got eaten somehow. I replaced the usage of CGLM_PI in the one file I missed. I also added all those other constants. Since none of them are in standard C, it makes sense to provide them in a math library like this.

include/cglm/types.h
```diff
@@ -48,8 +48,37 @@ typedef CGLM_ALIGN_IF(16) vec4  mat4[4];
...
+#define GLM_LN2f      ((float) GLM_LN2)
+#define GLM_LN10f     ((float) GLM_LN10)
+#define GLM_PIf       ((float) GLM_PI)
+#define GLM_PI_2f     ((float) GLM_PI_2)
+#define GLM_PI_4f     ((float) GLM_PI_4)
+#define GLM_1_PIf     ((float) GLM_1_PI)
+#define GLM_2_PIf     ((float) GLM_2_PI)
+#define GLM_2_SQRTPIf ((float) GLM_2_SQRTPI)
+#define GLM_SQRT2f    ((float) GLM_SQRT2)
+#define GLM_SQRT1_2f  ((float) GLM_SQRT1_2)
```
recp: style: no space between cast operator and value/macro pls. `((float) GLM_E)` -> `((float)GLM_E)` ... 

I'm very obsessive about style ☺️

recp:
@hartenfels merged, thanks for your contribution 🎉

Issue #62, opened 2018-10-04
Binary Constants in AVX implementations


jonathanplatzer:
While playing with various compiler flags on my machine with AVX support, I noticed that you are using binary constants in the `mm256_permute` functions, which are GCC extensions. This could lead to some problems when trying to compile on MSVC, but it is also not supported when using ISO C99. 

The question is now if we should just change these to hex constants or do something like this to make it more readable:
```c
#if _GNUC_
  s1 = 0b00000011;
#else
  s1 = 0x3;
#endif
y4 = _mm256_permute2f128_ps(y2, y2, s1);
```

recp:
@jonathanplatzer I didn't aware that binary constants are GCC extensions, thanks.

Maybe we can always use hex instead of if/else?

To make it more readable, I think we could add comments, for your case:
```C
y4 = _mm256_permute2f128_ps(y2, y2, 0x03); /* 0x03: 0b00000011 */
```

or a macro could help to make it more readable maybe. Because macro name could give us an idea what control/imm parameter does.

I'll read Intel Intrinsics Guide asap to remember control parameter (last parameter) usage. 

If you would like to fix it, a PR is welcome ;)

PR #63, opened 2018-10-12
avx: replace binary constants with hex

recp:
Replace binary constants (GCC extension) with hex (see https://github.com/recp/cglm/issues/62)


recp:
After run some tests, I realized that `glm_mul_avx()` fails in some cases, not always. (`glm_mat4_mul_avx()` passes in all tests). In my tests, first column of result matrix seems wrong others seem correct. I will fix it in this PR. 

After merged this PR, I will focus on ARM/Neon arch (there are some local changes to push).

PR #64, opened 2018-10-30
avx: AVX1 support for matrix inverse

recp:
**cglm** already supports AVX version for mat4_mul, but mat4_inv was missing. I implemented AVX1 version of matrix inverse. 

After upgraded my Macbook Pro I'll try to implement AVX2 + FMA too, but since my current CPU does not support that, I can't do that for now.

I tested mat4_inv on Ivy Bridge CPU, I got similar performance with SSE (not better), but on new CPUs the result may be different. I'll try to reduce some shuffles later to increase performance.

New functions:
- [x] `glm_mat4_scale_avx(mat4 m, float s)`
- [x] `glm_mat4_mul_avx(mat4 m1, mat4 m2, mat4 dest)`

These are selected automatically if -mavx is set. 

I'll try to optimize SIMD-ed functions with SSE3 and SSE4 later.

### DIFF
--- a/include/cglm/mat4.h
+++ b/include/cglm/mat4.h
@@ -528,7 +528,9 @@ glm_mat4_det(mat4 mat) {
 CGLM_INLINE
 void
 glm_mat4_inv(mat4 mat, mat4 dest) {
-#if defined( __SSE__ ) || defined( __SSE2__ )
+#ifdef __AVX__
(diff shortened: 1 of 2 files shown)


recp:
`glm_mat4_scale_avx()` is added to master and I'll try to re-implement the AVX vesion

PR #65, opened 2018-11-17
fix versor alignment

sinisterchipmunk:
[One of my unit tests](https://github.com/sinisterchipmunk/cglm-ruby/blob/96ec0f350051f9827f2f4aa995ef2f57e4f88def/test/cglm/quat_test.rb#L132) was crashing on the call to `glm_quat_inv`. I noticed the temporary versor inside of `glm_quat_inv` is aligned to `8` but other versors in the same file are aligned to `16`. I also see in your documentation that versors must be aligned to `16`. So I changed it. This seems to have resolved the crash for me.


recp:
@sinisterchipmunk good catch! It must be 16 not 8. 

The PR is merged, thanks for your contribution!

PR #79, opened 2019-03-29
Include `stddef.h` in `common.h`

acoto87:
Include `stddef.h` to ensure `size_t` and other dependent types.


recp:
@acoto87 merged 🎉, thanks for your contribution!

Issue #83, opened 2019-04-17
Functions don't use const when possible


kaadmy:
Due to functions not taking arguments as `const` when possible (ie. when a variable is never written to), it's very tough to integrate CGLM with code that passes vector/matrix types as `const`. Since constant variables cannot be cast to non-const, this means any code that calls CGLM can't use constant variables.

Since it's possible to cast *to* `const`, but not from, would it be possible to mark read-only function arguments as `const`?

recp:
Hi @kaadmy 

Do you use C++? Yes we could make readonly parameters `const` and `const` reference in C++ maybe to allow to pass initializer list as parameter

Something like this make that possible, I think:

`glm_mat4_mul(GLM_IN(mat4) a, GLM_IN(mat4) b, GLM_OUT(mat4) dest)`

it is more obvious which parameter is writable in this way

```C
#ifdef __cplusplus
#define GLM_IN(X)    const X& 
#define GLM_INOUT(X) const X& 
#define GLM_OUT(X)   const X
#else
#define GLM_IN(X)    const X
#define GLM_INOUT(X)       X
#define GLM_OUT(X)         X
#endif
```

kaadmy:
I'm currently using C99.

Only parameters that aren't written to need to be `const` (in fact, it's an error to use `const` otherwise), which includes nearly all functions that have a `src` and `dest` parameter.

recp:
I want to use `const X& ` in C++ to allow something like this:

```C
/* C and C++ version */
glm_vec3_norm((vec3){1, 2, 3});

/* C++ version if const X& is enabled otherwise: compiler error like 'No matching function for call to...' */
glm_vec3_norm({1, 2, 3});
```

I think `GLM_IN`, `GLM_OUT`, `GLM_INOUT` macros will make **C** and **C++** happy

heapseeker:
@recp The issue isn't C++ or not C++. The issue is that read-only variables that aren't mutated are not marked as const, which they should be. I may make a PR and do this, as it makes my own code a bit sloppy because I can't use const. 

Also I would recommend **against** caring about C++. Realisitically no one would use this library over GLM if you're using C++. GLM is just a more mature, stable library. Keep your focus and energy on everything C related.

recp:
> Also I would recommend against caring about or supporting C++. Realisitically no one would use this library over GLM if you're using C++. GLM is just a more mature, stable library. Keep your focus and energy on everything C related.

Fair enough.

[...]

PR #86, opened 2019-04-28
mark readonly parameters as const

recp:
see https://github.com/recp/cglm/issues/83

### DIFF
--- a/include/cglm/affine-mat.h
+++ b/include/cglm/affine-mat.h
@@ -44,7 +44,7 @@
  */
 CGLM_INLINE
 void
-glm_mul(mat4 m1, mat4 m2, mat4 dest) {
+glm_mul(const mat4 m1, const mat4 m2, mat4 dest) {
 #ifdef __AVX__
   glm_mul_avx(m1, m2, dest);
 #elif defined( __SSE__ ) || defined( __SSE2__ )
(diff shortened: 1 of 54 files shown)


acoto87:
I think it looks good. It's true that parameters that aren´t modified should be `const` to allow library users have `const` if they want.

How would this affect the struct implementation? Should struct parameters that aren´t modified be `const` as well?

recp:
@acoto87 thanks for the review. 

> How would this affect the struct implementation? Should struct parameters that aren´t modified be const as well?

Yes I think it should be `const`, to prevent struct be copied in C++, we could do something like this:


```C
CGLM_INLINE
mat4s
glm_mat4_mul(const mat4s CPP_REF a, const mat4s CPP_REF b)
```

it is similar to my comment (https://github.com/recp/cglm/issues/83#issuecomment-484377636) but alternative.

```C
#ifdef __cplusplus
#define CPP_REF &
#else
#define CPP_REF
#endif
```

or maybe we keep focus on C and do this (C++ ref) in the future if needed

acoto87:
Well it's your call if you want to support C++ features, but I think it would be better at the moment to keep it simple.

> Yes I think it should be const, to prevent struct be copied in C++, we could do something like this:

Prevent copies when struct arguments are passed to functions is something you want to accomplish? In C the only way we have to do that is with pointers, right? Is that something you want to consider?

recp:
It would be better to not copy structs if it is possible (it is possible in C++), since our structs are inline maybe compiler may not copy them, I'm not sure for certain. Personally I don't want to use pointers in struct params yet. Should it be pointer?

Let's keep it simple as you said for now and ignore what I said for C++, in the future we may support to move/ref if we want, it is not topic of today.

PR #89, opened 2019-06-01
Make Building With Prefix Work

hartenfels:
Context: I'm using a configure prefix to build native and emscripten versions of this library in different directories. In that case, the current directory will be the build directory, not the source directory.

Steps to reproduce:

* Create a directory and cd into it.
* `/path/to/cglm/configure --prefix="$(pwd)"`
* `make`

Without this patch, it'll try to run `./post-build.sh`, which fails cause it's not in `.`.

With this patch, it'll run `/path/to/cglm/post-build.sh` instead, which is the right thing to do.

### DIFF
--- a/makefile.am
+++ b/makefile.am
@@ -132,5 +132,11 @@ test_tests_SOURCES=\
     test/src/test_affine.c \
     test/src/test_bezier.c
 
+# When running configure with --prefix, $VPATH references
+# the source directory that post-build.sh is in. When not
+# using a prefix, $VPATH will be unset, so we need to fall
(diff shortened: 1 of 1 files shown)


recp:
The make file seems to be outdated, I'll try to update it asap. If you or someone else want to fix that, you are welcome!

It is outdated because, it does not contain all new headers. **cglm** now supports Struct API as alternative syntax to existing Array API. Make file and Windows build file need to be updated for this update. I was missed this.

Anyway, thank you for fixing this and your contribution! It is merged now 🎉

PR #90, opened 2019-06-23
Add pkg-config Support + Autotools Fixes

hartenfels:
Adds a pkg-config file, making dependency resolution easier. After this, you can run `pkg-config --cflags cglm` and `pkg-config --libs cglm` and get the compiler and linker flags for cglm. Other build systems use this, like autotools and Meson.

There's some wonkiness with autotools installing to `/usr/local/lib/pkgconfig` by default, which isn't necessarily a search path for it. That's documented in the README.

Along the way I also added a pair of missing files and consistenized the Makefile capitalization, they were different in the .gitignore and the autotools config and hence not getting ignored properly.


hartenfels:
Hrm, okay, apparently Travis CI has a really, really old version of pkg-config from like 2012, which doesn't have the `PKG_INSTALLDIR` macro yet. In that case it now gets emulated manually and the checks are all green.

recp:
@hartenfels this PR is merged now, thank you for your contributions 🎉

PR #94, opened 2019-08-21
new name for euler sequence

superbigio:


include/cglm/euler.h
```diff
@@ -55,6 +55,8 @@ typedef enum glm_euler_sq {
   GLM_EULER_ZYX = 2 << 0 | 1 << 2 | 0 << 4
 } glm_euler_sq;
 
+typedef glm_euler_sq glm_euler_seq;
+
```
recp: this defines new name but doesn't change existing usage. I think functions that use `glm_euler_sq` should use `glm_euler_seq` instead.

superbigio: Yes, sorry about that. Your suggestion makes more sense. I made the appropriate changes however I left the typedef in so that people using the old version of the name (glm_euler_sq) could still be able to compile. New PR for you to review.

recp:
@luigi @superbigio thanks for your contribution! The PR is merged 🎉

PR #96, opened 2019-08-22
squared distance with SIMD support

superbigio:
PR

### DIFF
--- a/include/cglm/vec3.h
+++ b/include/cglm/vec3.h
@@ -48,14 +48,14 @@
    CGLM_INLINE void  glm_vec3_normalize_to(vec3 v, vec3 dest);
    CGLM_INLINE void  glm_vec3_cross(vec3 a, vec3 b, vec3 d);
    CGLM_INLINE void  glm_vec3_crossn(vec3 a, vec3 b, vec3 dest);
-   CGLM_INLINE float glm_vec3_distance(vec3 a, vec3 b);
    CGLM_INLINE float glm_vec3_angle(vec3 a, vec3 b);
    CGLM_INLINE void  glm_vec3_rotate(vec3 v, float angle, vec3 axis);
    CGLM_INLINE void  glm_vec3_rotate_m4(mat4 m, vec3 v, vec3 dest);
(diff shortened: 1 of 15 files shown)


include/cglm/vec3.h
```diff
@@ -694,9 +694,9 @@ glm_vec3_center(vec3 a, vec3 b, vec3 dest) {
 CGLM_INLINE
 float
 glm_vec3_distance2(vec3 a, vec3 b) {
-  return glm_pow2(b[0] - a[0])
-       + glm_pow2(b[1] - a[1])
-       + glm_pow2(b[2] - a[2]);
+  return glm_pow2(a[0] - b[0])
+       + glm_pow2(a[1] - b[1])
+       + glm_pow2(a[2] - b[2]);
```
recp: May I ask what is the benefit[s] of using `a` first instead of b?

include/cglm/vec4.h
```diff
@@ -697,14 +698,36 @@ CGLM_INLINE
 float
 glm_vec4_distance(vec4 a, vec4 b) {
 #if defined( __SSE__ ) || defined( __SSE2__ )
-  return glmm_norm(_mm_sub_ps(glmm_load(b), glmm_load(a)));
+  return glmm_norm(_mm_sub_ps(glmm_load(a), glmm_load(b)));
```
recp: May I ask what is the benefit[s] of using a first instead of b?

include/cglm/vec3.h
recp: I used `b - a` here because the distance formula is like `sqtr(pow(x2 - x1, 2) + pow(y2 - y1, 2) + ...)`

include/cglm/vec4.h
superbigio: There is no real benefit. It is absolutely the same because the subtraction is then raised to the 2nd power, so the result will always be positive. I just like to see a - b because that's the natural order of the function arguments and there is no reason to put b before a. Inverting the natural order suggests that there is a reason why b needs to come before a, which is not the case.

recp: Fair enough.

recp:
@superbigio thanks for your contribution!

Issue #15, opened 2018-01-03
TODOs and Features Queue


recp:
General
===
- [ ] Unit tests (In Progress) **[important]**
- [ ] ARM Neon Arch (In Progress)
- [ ] Unit tests for comparing cglm with glm results **[important]**
- [ ] Unaligned operations (e.g. glm_umat4_mul)
- [ ] Extra documentation
- [ ] Performance tests

Quaternion
===
- [ ] Matrix to Quaternion
- [ ] Rotate vector with quat (we already have `glm_vec_rotate(vec, angle, axis)`)
- [ ] Half precision float (16 bit) support
- [ ] Unit tests

Precision
===
- [ ] Half precision float (16 bit) 
- [ ] Double precision float (64 bit double) 

Convenient Functions
===
There could be inline wrapper for some functions e.g. `glm_vec_cross`and `glm_vec_dot` like `glm_cross` and `glm_dot`. Original functions must stay where they are. This is not necessary but could help to write code fast and more readable . 

One another things there could be convenient functions for operations like this ( it is just example ):
```C
/* v = v1 * v2 + v3 * v4 + v5 * v6 */
vec3 sum;
vec3 tmp, tmp2;

glm_vec_mulv(v1, v2, sum);

glm_vec_mulv(v3, v4, tmp2);
glm_vec_add(tmp2, sum, sum);

glm_vec_mulv(v5, v6, tmp2);
glm_vec_add(tmp2, sum, sum);
/* ... */
```

It could be something like this:
```C
/* v = v1 * v2 + v3 * v4 + v5 * v6 */
vec3 sum;

glm_vec_mulv(v1, v2, sum);

glm_vec_muladd(v3, v4, sum);
glm_vec_muladd(v5, v6, sum);

/* ... */
```

this is also not necessary, just my thoughts. In short I mean convenient functions for +=, *=, -=, /=

Also any feature request, feedback or contribution are welcome. We can use this issue collect feature requests and feedbacks

recp:
Rename `glm_scale1` to `glm_scale_uni`

recp:
Find perpendicular vector util e.g. `glm_vec_perpendicular` or `glm_vec_orthogonal` or `glm_vec_ortho`, the last one seems cool! 

I found good an fast solution, only subtraction: https://www.quora.com/How-do-I-find-a-vector-perpendicular-to-another-vector/answer/Tom-Thompson-1?srid=uEqzq

[...]

PR #98, opened 2019-08-24
glm_lerpc(), glm_step(), glm_smoothstep(), glm_smoothinterp()

superbigio:
PR implementing step and smoothstep functions (following the OpenCL specs) as outlined in #15. In addition glm_lerp() has been made non-clamping and glm_lerpc() has been created to have a clamping version and to be more flexible and consistent with glm_percent/glm_percentc.
A glm_smoothinterp() function is also been implemented as a natural extension to the smoothstep interpolating equation.

A few points worth discussing:
1) glm_step() has 2 different implementations (branching or type conversion). Could not test which one would be better/more efficient. Need some help with that.
2) The argument order of glm_step() and glm_smoothstep() is non consistent with the rest of the API. However I implemented it that way to stay consistent with the versions outlined in the OpenCL specs. I am not sure which argument order I should follow. Suggestions welcome.
3) Anything else?

### DIFF
--- a/include/cglm/util.h
+++ b/include/cglm/util.h
@@ -213,11 +213,11 @@ float
 glm_step(float edge, float x)
 {
 #if 1
-    /* type conversion - no branching */
-    return !(x < edge);
+  /* type conversion - no branching */
+  return !(x < edge);
 #else
-    /* branching - no type conversion */
-    return (x < edge) ? 0.0f : 1.0f;
+  /* branching - no type conversion */
+  return (x < edge) ? 0.0f : 1.0f;
 #endif
 }
 
--- a/include/cglm/util.h
+++ b/include/cglm/util.h
@@ -202,7 +202,7 @@ glm_lerpc(float from, float to, float t) {
 }
 
 /*!
- * @brief threshold function (according to OpenCL specs)
+ * @brief threshold function
  *
  * @param[in]   edge    threshold
  * @param[in]   x       value to test against threshold
@@ -214,7 +214,8 @@ glm_step(float edge, float x) {
   /* branching - no type conversion */
   return (x < edge) ? 0.0f : 1.0f;
   /*
-   * An alternative implementation could be:
+   * An alternative implementation without branching
+   * but with type conversion could be:
    * return !(x < edge);
    */
 }
--- a/include/cglm/util.h
+++ b/include/cglm/util.h
@@ -210,15 +210,13 @@ glm_lerpc(float from, float to, float t) {
  */
 CGLM_INLINE
 float
-glm_step(float edge, float x)
-{
-#if 1
-  /* type conversion - no branching */
-  return !(x < edge);
-#else
+glm_step(float edge, float x) {
   /* branching - no type conversion */
   return (x < edge) ? 0.0f : 1.0f;
-#endif
+  /*
+   * An alternative implementation could be:
+   * return !(x < edge);
+   */
 }
 
 /*!
--- a/include/cglm/util.h
+++ b/include/cglm/util.h
@@ -8,11 +8,26 @@
 /*
  Functions:
    CGLM_INLINE int   glm_sign(int val);
+   CGLM_INLINE float glm_signf(float val);
    CGLM_INLINE float glm_rad(float deg);
    CGLM_INLINE float glm_deg(float rad);
    CGLM_INLINE void  glm_make_rad(float *deg);
    CGLM_INLINE void  glm_make_deg(float *rad);
    CGLM_INLINE float glm_pow2(float x);
+   CGLM_INLINE float glm_min(float a, float b);
+   CGLM_INLINE float glm_max(float a, float b);
+   CGLM_INLINE float glm_clamp(float val, float minVal, float maxVal);
+   CGLM_INLINE float glm_clamp_zo(float val, float minVal, float maxVal);
+   CGLM_INLINE float glm_lerp(float from, float to, float t);
+   CGLM_INLINE float glm_lerpc(float from, float to, float t);
+   CGLM_INLINE float glm_step(float edge, float x);
+   CGLM_INLINE float glm_smooth(float t);
+   CGLM_INLINE float glm_smoothstep(float edge0, float edge1, float x);
+   CGLM_INLINE float glm_smoothinterp(float from, float to, float t);
+   CGLM_INLINE float glm_smoothinterpc(float from, float to, float t);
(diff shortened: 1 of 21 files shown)


include/cglm/util.h
```diff
@@ -157,18 +172,113 @@ glm_clamp_zo(float val) {
...
  * @param[in]   from from value
  * @param[in]   to   to value
- * @param[in]   t    interpolant (amount) clamped between 0 and 1
+ * @param[in]   t    interpolant (amount)
  */
 CGLM_INLINE
 float
 glm_lerp(float from, float to, float t) {
-  return from + glm_clamp_zo(t) * (to - from);
+  return from + t * (to - from);
```
recp: The `t` parameter is clamped in all lerp functions (single float, vec4, vec3...). This breaks existing api consistency. I'm not sure if we should break existing api. On the other hand this will improve speed a bit which is good.

Options:
1. We could provide unclamped version which will not break api with another name? Like `glm_lerp_uc` **uc** suffix stands for un-clamped? Or you may come with better name
2. We could remove clamp from vec3 and vec4 and quat... too

Also it seems GLSL's mix functions don't apply clamping 🤔

include/cglm/util.h
```diff
@@ -157,18 +172,113 @@ glm_clamp_zo(float val) {
...
+ * @brief threshold function (according to OpenCL specs)
+ *
+ * @param[in]   edge    threshold
+ * @param[in]   x       value to test against threshold
+ * @return      returns 0.0 if x < edge, else 1.0
+ */
+CGLM_INLINE
+float
+glm_step(float edge, float x)
+{
```
recp: no newline for `{` pls.

recp:
The result of `!(x < edge)` is boolean while the return type is float32. In this case there is no jump but there is **bool/int-to-float** overhead. I don't now which one is better for now.

To eliminate type conversion let's continue with `(x < edge) ? 0.0f : 1.0f` until we are sure that `!(x < edge)` is fastest way to go.

include/cglm/vec3.h
```diff
@@ -793,6 +796,59 @@ glm_vec3_lerp(vec3 from, vec3 to, float t, vec3 dest) {
...
+ * @brief threshold function (according to OpenCL specs)
+ *
+ * @param[in]   edge    threshold
+ * @param[in]   x       value to test against threshold
+ * @param[out]  dest    destination
+ */
+CGLM_INLINE
+void
+glm_vec3_step(float edge, vec3 x, vec3 dest) {
+  dest[0] = glm_step(edge, x[0]);
```
recp: https://thebookofshaders.com/glossary/?search=step

You only provided edge as single dimension. it's size could be same as vector? We could provide `glm_vec3_step_uni` or `glm_vec3_step1`... for single dimension. Or `glm_vec3_stepv3` or `glm_vec3_step3` could be used for vec3?

recp:
> The argument order of glm_step() and glm_smoothstep() is non consistent with the rest of the API.

If the result can be stored elsewhere than source, then the dest is last parameter. Otherwise there is no explicit **dest** param and first parameter is **source** and **dest**.

include/cglm/util.h
superbigio: Sorry about that. Fixed.

include/cglm/util.h
superbigio: Yes, the GLSL specs do not apply clamping so I think the glm functions in order to have maximum consistency with GLSL should not apply clamping either. For example I would adopt a scheme such as the following: glm_lerp() [no clamping/standard] and then glm_lerpc() [with clamping]. This way we achieve maximum flexibility/efficiency by leaving it up to the user to pick the appropriate function (clamped but slower or non-clamped but faster). Same goes to vec3, vec4 and quat. So I would go with option 2. Existing users will have to revise their code.

As far as GLSL's mix goes: I have the implementation of a glm_mix() function, however I found it too similar to glm_lerp() to justify adding it and I ended up leaving it out. If you want, it would take me 5 mins to add it to a new PR.

include/cglm/vec3.h
superbigio: Yes, what you are saying is true.
I will change the single dimension case to glm_vec3_step_uni() and I will add vector versions.
Same goes for smoothstep.

include/cglm/util.h
recp: Ok then, let's go with 2.

> I found it too similar to glm_lerp() 

Actually it must be same, just different name convention.

`glm_mix` is really cool name rather than `glm_lerp`, especially for colors. Maybe we can provide both (`mix` as wrapper of `lerp`), so GLSL/Renderman users would use **mix** and HLSL users would use **lerp**.

superbigio: ok, sounds good to me.

superbigio: As lerp goes, now we also have glm_lerpc() which is the clamped version of glm_lerp(). Should I make a clamped version of glm_mix() as well? So in the end we would have glm_mix() wrapping glm_lerp() [non-clamped] and then glm_mixc() wrapping glm_lerpc() [clamped].

recp: yes, but I think `glm_mixc()` must be inline like `glm_mix()` because both are wrappers.

superbigio: Ok, I updated the PR.

include/cglm/vec4.h
```diff
@@ -815,12 +822,164 @@ glm_vec4_lerp(vec4 from, vec4 to, float t, vec4 dest) {
...
+ *
+ * @param[in]   edge0   low threshold
+ * @param[in]   edge1   high threshold
+ * @param[in]   x       value to test against threshold
+ * @param[out]  dest    destination
+ */
+CGLM_INLINE
+void
+glm_vec4_smoothstep(vec4 edge0, vec4 edge1, vec4 x, vec4 dest) {
+  dest[0] = glm_smoothstep(edge0[0], edge1[0], x[0]);
```
recp: I think `x` must be first parameter, what you you think?

include/cglm/vec3.h
recp: I think `x` must be first parameter, what you you think?

include/cglm/vec4.h
recp: Forget it, the current version is same as GLSL which is may be fine.

recp:
@superbigio this PR is merged now, thanks for your contributions 🎉

PR #103, opened 2019-08-27
some useful functions added

superbigio:
List of functions added:
abs()
fract()
norm_one()
norm_inf()
hadd()
hmax() [only in x86]


include/cglm/simd/x86.h
```diff
@@ -68,6 +74,38 @@ glmm_hadd(__m128 v) {
...
+  return _mm_cvtss_f32(glmm_vhmin(v));
+}
+
+static inline
+__m128
+glmm_vhmax(__m128 v) {
+  __m128 x0, x1, x2;
+  x0 = _mm_movehl_ps(v, v); /* [2, 3, 2, 3] */
+  x1 = _mm_max_ps(x0, v); /* [0|2, 1|3, 2|2, 3|3] */
+  x2 = _mm_shuffle_ps(x1, x1, _MM_SHUFFLE(1, 1, 1, 1)); /* [1|3, 1|3, 1|3, 1|3] */
```
recp: `glmm_shuff1x` could be used here:

```C
x2 = glmm_shuff1x(x1, 1);
```

include/cglm/vec3-ext.h
```diff
@@ -211,6 +214,46 @@ glm_vec3_sign(vec3 v, vec3 dest) {
...
+
+/*!
+ * @brief vector reduction by summation
+ *
+ * @param[in]  v    vector
+ * @return     sum of all vector's elements
+ */
+CGLM_INLINE
+float
+glm_vec3_hadd(vec3 v) {   // could overflow
```
recp: no `C++/C99+` style comments, use C89 comments pls.

include/cglm/simd/x86.h
```diff
@@ -68,6 +74,38 @@ glmm_hadd(__m128 v) {
   return _mm_cvtss_f32(glmm_vhadds(v));
 }
 
+static inline
+__m128
+glmm_vhmin(__m128 v) {
+  __m128 x0, x1, x2;
+  x0 = _mm_movehl_ps(v, v); /* [2, 3, 2, 3] */
+  x1 = _mm_min_ps(x0, v); /* [0|2, 1|3, 2|2, 3|3] */
+  x2 = _mm_shuffle_ps(x1, x1, _MM_SHUFFLE(1, 1, 1, 1)); /* [1|3, 1|3, 1|3, 1|3] */
```
recp: same as below.

include/cglm/vec4.h
```diff
@@ -256,6 +258,44 @@ glm_vec4_norm(vec4 v) {
...
+ * @brief infinity norm of vec4
+ *
+ * @param[in] v vector
+ *
+ * @return infinity norm
+ */
+CGLM_INLINE
+float
+glm_vec4_norm_inf(vec4 v) {
+#if defined(CGLM_SIMD)
```
recp: `CGLM_SIMD` is defined when one of SIMD is supported. If NEON_FP is present then this will (or should) be defined. But it seems `glmm_norm_inf()` is only defined for x86. In this case it you should test SIMD against `CGLM_SIMD_x86` not `CGLM_SIMD`, otherwise it cannot be compiled on ARM I guess.

include/cglm/vec4.h
```diff
@@ -256,6 +258,44 @@ glm_vec4_norm(vec4 v) {
...
+ *
+ * @param[in] v vector
+ *
+ * @return L1 norm
+ */
+CGLM_INLINE
+float
+glm_vec4_norm_one(vec4 v) {
+#if defined(CGLM_SIMD)
+  return glmm_norm_one(glmm_load(v));
```
recp: `CGLM_SIMD` is defined when one of SIMD is supported. If NEON_FP is present then this will (or should) be defined. But it seems `glmm_norm_one()` is only defined for x86. In this case it you should test SIMD against `CGLM_SIMD_x86` not `CGLM_SIMD`, otherwise it cannot be compiled on ARM I guess.

include/cglm/vec4-ext.h
```diff
@@ -237,6 +240,58 @@ glm_vec4_sign(vec4 v, vec4 dest) {
...
+
+/*!
+ * @brief vector reduction by summation
+ *
+ * @param[in]  v    vector
+ * @return     sum of all vector's elements
+ */
+CGLM_INLINE
+float
+glm_vec4_hadd(vec4 v) {   /* could overflow */
```
recp: Could you move the comment below of brief (with extra new line)? I think that is better place for comments.

include/cglm/vec4.h
superbigio: Yes, you are absolutely right. I will try to implement `glmm_norm_inf()` for ARM and keep the `CGLM_SIMD` definition. Otherwise I will switch to `CGLM_SIMD_x86`, as you said.

include/cglm/vec4.h
superbigio: Yes, you are absolutely right. I will try to implement `glmm_norm_one()` for ARM and keep the `CGLM_SIMD` definition. Otherwise I will switch to `CGLM_SIMD_x86`, as you said.

recp: Can we add more detail to brief of these functions;

Some resources:
https://medium.com/@montjoile/l0-norm-l1-norm-l2-norm-l-infinity-norm-7a7d18a4f40c
https://machinelearningmastery.com/vector-norms-machine-learning/

Also adding new functions to documentation would be awesome if you have enough time 🤗 You can find `docs` folder, cglm uses **sphinx** for documentation.

superbigio: Hi Recep, I will definitely add more info to the brief of the two norm functions. That's no problem. However, I am currently working on two projects at once and I don't have time to update the documentation.

recp: > I don't have time to update the documentation.

No problem, I'll update it asap.

recp:
Some tests would be awesome, but there are many things to test; so we can postpone it for now. 

Test cases are welcome for any functions of course 🤗

After we made tests more stable then all Pull Requests or features must come with test cases (if the changes are related to coding)

recp:
@superbigio thanks for your contributions 🎉

PR #109, opened 2019-11-23
Change `()` Prototypes to `(void)`

hartenfels:
There's a few functions that have an empty prototype `()`, which should be `(void)` instead. This patch fixes that.

Since it's nonsense to have an empty prototype, I've added `-Werror=strict-prototypes` to the build options. This will throw a compile error in these cases so they can be avoided ahead of time.


recp:
@hartenfels thanks for your contribution 🎉

PR #111, opened 2019-11-24
Fix Automake Flags and Matrix Struct Initializers

hartenfels:
The automake configuration was building things with `-O3 -O2`, because for some reason `-O2` is a default flag. This PR removes the default flags so that we build *only* with `-O3`.

It was also building the tests entirely without warnings, which silenced some mistakes. That's been fixed and warnings in tests aren't being suppressed anymore. This revealed two errors:

They complained about the anonymous structures, so the `-std=gnu99` flag was changed to `-std=gnu11`, since anonymous structs were introduced in C11.

It also didn't like the matrix struct initializers, since they were missing a set of braces around them. One more set of braces to fix that one.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -151,7 +151,8 @@ test_tests_SOURCES=\
     test/src/test_vec3.c \
     test/src/test_mat3.c \
     test/src/test_affine.c \
-    test/src/test_bezier.c
+    test/src/test_bezier.c \
+    test/src/test_struct.c
 
 pkgconfig_DATA=cglm.pc
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -21,7 +21,8 @@ libcglm_la_LDFLAGS = -no-undefined -version-info 0:1:0
 checkLDFLAGS = -L./.libs \
                -lm \
                -lcglm
-checkCFLAGS = -I./include
+checkCFLAGS = $(AM_CFLAGS) \
+              -I./include
 
 check_PROGRAMS = test/tests
 TESTS = $(check_PROGRAMS)
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,7 +8,7 @@
 
 ACLOCAL_AMFLAGS = -I m4
 AM_CFLAGS = -Wall \
-            -std=gnu99 \
+            -std=gnu11 \
             -O3 \
             -Wstrict-aliasing=2 \
             -fstrict-aliasing \
--- a/configure.ac
+++ b/configure.ac
@@ -10,6 +10,9 @@ AC_PREREQ([2.69])
 AC_INIT([cglm], [0.6.1], [info@recp.me])
 AM_INIT_AUTOMAKE([-Wall -Werror foreign subdir-objects serial-tests])
 
+# Don't use the default cflags (-O2 -g), we set ours manually in Makefile.am.
+: ${CFLAGS=""}
+
 AC_CONFIG_MACRO_DIR([m4])
 AC_CONFIG_SRCDIR([src/])
 AC_CONFIG_HEADERS([config.h])
--- a/include/cglm/struct/mat3.h
+++ b/include/cglm/struct/mat3.h
@@ -38,12 +38,8 @@
 #include "../mat3.h"
 #include "vec3.h"
 
-#define GLMS_MAT3_IDENTITY_INIT  {{1.0f, 0.0f, 0.0f,                         \
-                                   0.0f, 1.0f, 0.0f,                         \
-                                   0.0f, 0.0f, 1.0f}}
-#define GLMS_MAT3_ZERO_INIT      {{0.0f, 0.0f, 0.0f,                         \
-                                   0.0f, 0.0f, 0.0f,                         \
-                                   0.0f, 0.0f, 0.0f}}
+#define GLMS_MAT3_IDENTITY_INIT  {GLM_MAT3_IDENTITY_INIT}
+#define GLMS_MAT3_ZERO_INIT      {GLM_MAT3_ZERO_INIT}
 
 /* for C only */
 #define GLMS_MAT3_IDENTITY ((mat3s)GLMS_MAT3_IDENTITY_INIT)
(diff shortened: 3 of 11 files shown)


Makefile.am
```diff
@@ -8,7 +8,7 @@
 
 ACLOCAL_AMFLAGS = -I m4
 AM_CFLAGS = -Wall \
-            -std=gnu99 \
+            -std=gnu11 \
```
recp: I think we should not change this right now (in the future, maybe), because **cglm** should work with C89 and C99. Upgrading C version may cause to ignore some warnings for old versions. We cannot suppress warnings by upgrading C version. 

Removing `-pedantic` would also suppress warnings if there is any about structs.

configure.ac
```diff
@@ -10,6 +10,9 @@ AC_PREREQ([2.69])
 AC_INIT([cglm], [0.6.1], [info@recp.me])
 AM_INIT_AUTOMAKE([-Wall -Werror foreign subdir-objects serial-tests])
 
+# Don't use the default cflags (-O2 -g), we set ours manually in Makefile.am.
+: ${CFLAGS=""}
```
recp: Is there any other default flags? We are removing them all, is there any side effect to remove them entirely?

include/cglm/struct/mat3.h
```diff
@@ -38,12 +38,12 @@
...
 
-#define GLMS_MAT3_IDENTITY_INIT  {1.0f, 0.0f, 0.0f,                          \
-                                  0.0f, 1.0f, 0.0f,                          \
-                                  0.0f, 0.0f, 1.0f}
-#define GLMS_MAT3_ZERO_INIT      {0.0f, 0.0f, 0.0f,                          \
-                                  0.0f, 0.0f, 0.0f,                          \
-                                  0.0f, 0.0f, 0.0f}
+#define GLMS_MAT3_IDENTITY_INIT  {{1.0f, 0.0f, 0.0f,                         \
+                                   0.0f, 1.0f, 0.0f,                         \
+                                   0.0f, 0.0f, 1.0f}}
```
recp: Maybe we could use matrix initializers like this instead:

```C
#define GLMS_MAT3_IDENTITY_INIT  {{1.0f, 0.0f, 0.0f},                         \
                                  {0.0f, 1.0f, 0.0f},                         \
                                  {0.0f, 0.0f, 1.0f}}
```

I have checked the array api and they are initialized like this. Maybe we could use the same

Makefile.am
hartenfels: It works just fine with C99, but you need to build it with `-DCGLM_NO_ANONYMOUS_STRUCT`. Anonymous structures are a C11 feature, so cglm is already using it. Pretending to build in C99 mode while using a C11 feature is strange.

A _user_ of the library can always use C99. They just `#define CGLM_NO_ANONYMOUS_STRUCT` before `#include`ing the CGLM headers. So even if we build with C11, we don't break any compatibility.

Maybe this can just be added to the README? I can add a section that explains all this and add it to this PR.

Please don't remove `-pedantic` though, that's just gonna push the warnings to the user.

configure.ac
hartenfels: There's no other default flags. The Autoconf documentation recommends this: <https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/C-Compiler.html#index-AC_005fPROG_005fCC-861>

> If output variable `CFLAGS` was not already set, set it to `-g -O2` for the GNU C compiler (`-O2` on systems where GCC does not accept `-g`), or `-g` for other compilers. If your package does not like this default, then it is acceptable to insert the line `: ${CFLAGS=""}` after `AC_INIT` and before `AC_PROG_CC` to select an empty default instead.

Since we don't want `-g` or `-O2`, an empty default makes more sense. The user can still override this by setting the `CFLAGS` variable explicitly.

include/cglm/struct/mat3.h
hartenfels: No, that doesn't work. It picks the wrong field to initialize. The only thing that works is a designated initializer:

```c
#define GLMS_MAT3_IDENTITY_INIT  {.col = {{1.0f, 0.0f, 0.0f},                \
                                          {0.0f, 1.0f, 0.0f},                \
                                          {0.0f, 0.0f, 1.0f}}}
```

But that's C99-only and I think MSVC doesn't support it.

Makefile.am
recp: Okay,


>  Anonymous structures are a C11 feature, so cglm is already using it. Pretending to build in C99 mode while using a C11 feature is strange.


I think it can be used without C11, with language extensions, so "pretending" is not really true here. If compiler uses C89/C99 and doesn't have any extension about anonymous structs then user can define `CGLM_NO_ANONYMOUS_STRUCT` to disable it. 

But switching to C11 seems not to be bad idea. Let's do it.

Another option, maybe we could check compiler features, version... to enable/disable it automatically ?

> Maybe this can just be added to the README? I can add a section that explains all this and add it to this PR.

Please.

Contribution about README, Documentation, Tests are also welcome ;)

> Please don't remove -pedantic though, that's just gonna push the warnings to the user.

Let's keep it.

include/cglm/struct/mat3.h
recp: > No, that doesn't work. It picks the wrong field to initialize.

May I ask why wrong? We are initializing the whole struct at once. 

> But that's C99-only and I think MSVC doesn't support it.

Agree. Let's not use designated initializers.

hartenfels: I think I figured it out. It was picking the anonymous struct over `col` because that's what comes first. Just re-ordering the `union` so that `col` comes first fixes it. I'll amend this PR.

Makefile.am
hartenfels: > Another option, maybe we could check compiler features, version... to enable/disable it automatically ?

I can implement that. It involves a bit of a dance to stay backward-compatible and such, but it should be possible. I'll be amending the PR as such.

hartenfels:
I've made the anonymous struct enabling/disabling automatic and documented the behavior in the README. The struct initializers now all work properly without warnings and simply use the array initializers under the hood, which is less repetition.

recp:
@hartenfels really thanks!

Issue #114, opened 2020-01-14
Build error on Android


kinnefix:
I am trying to use this library for my project.
On running make, I get the following errors:
[output.log](https://github.com/recp/cglm/files/4058435/output.log)

kinnefix:
And of course, the build succeeds if I make a modification to file include/cglm/simd/intrin.h to disable definition of CGLM_NEON_FP.

recp:
Hello @kinnefix 

Thanks for reporting this issue 

It seems there are type casting errors (`float32x4_t` to `float32x2_t`) this can be fix by creating temp variables for them instead of assigning them to float34x4_t...

I will fix it asap but if you want to fix it before me a PR is welcome ;)

recp:
@kinnefix can you try to build using the PR: https://github.com/recp/cglm/pull/115

PR #115, opened 2020-01-15
build fix for ARM NEON

recp:
fixes: https://github.com/recp/cglm/issues/114


kinnefix:
Just confirmed that it works!
Sorry for the late reply though.

recp:
@kinnefix thanks 🤗

PR #121, opened 2020-03-03
fix glms_sphere_transform by using mat4s instead of mat4

Zollerboy1:
Currently the struct version of the `gem_sphere_transform()` function does take a `vec4s` as the first parameter. Strangely enough the second parameter is not a `mat4s` as expected, but rather the normal `mat4`.

Because this is the only function in the whole `glms`-API, which takes a non-struct parameter, it has to be a mistake.


recp:
@Zollerboy1 thanks for your contribution 🎉

Zollerboy1:
You’re welcome 😊 
I’m currently building a Swift wrapper around cglm and without the struct-API this would be an absolute nightmare.

recp:
Awesome !

I could use those wrappers for myself ;) You can also check [cglm/applesimd.h](https://github.com/recp/cglm/blob/master/include/cglm/applesimd.h) header. I have created that header to translate **cglm** types to Apple SIMD library types to send **cglm** types to Metal GL without alignment issues. Apple/ObjC/Swift compatibility is very important

PR #122, opened 2020-03-04
fix glms_vec2_rotate() by using a dest vec2s instead of taking an axis vector

Zollerboy1:
The `glms_vec2_rotate()` function was taking an axis `vec2s` and passed it to the `glm_vec2_rotate()` function, which does not even need an axis vector, but rather a dest vector.

Therefore the axis vector just got overridden by the result of the rotation. Because the `glms_vec2_rotate()` function returned the unchanged vector there was no possibility, to get the result of the function.


recp:
Many thanks @Zollerboy1, the PR is merged 🎉

PR #125, opened 2020-03-11
Fix small documentation mistake

Uwila:
I think I noticed a small mistake in the documentation: At two points in the documentation the names of the parameters in the parameters list do not match the parameter names in the function:
- `glm_look`
- `glm_look_anyup`

Namely the parameter `center` which should I think should be `dir`.


recp:
Thanks @Uwila

PR #126, opened 2020-03-20
Fix documentation mistake for `glm_vec3_rotate`

Uwila:
I think I found another small mistake in the documentation; This, for `glm_vec3_rotate`, correctly labels `angle` as `in` rather than `out`.


recp:
Thanks for your contribution 🎉

PR #131, opened 2020-04-02
Add ray-triangle intersection check

### LINKED ISSUE #129: Möller–Trumbore intersection algorithm
Perhaps the [Möller–Trumbore intersection algorithm](https://en.wikipedia.org/wiki/M%C3%B6ller%E2%80%93Trumbore_intersection_algorithm) would make a good addition to this library? I would be willing to try and implement it; I already made a version of it for a personal project that just uses C and cglm, so it wouldn't be too hard.

I'm not sure if you want a contribution like this though, and I also don't know whether it would be compatible with the license. The Wikipedia page is CC-SA 3.0 (I don't know if in this case a simple notice in a comment would suffice to point out the source, [this is what Wikipedia seems to imply though](https://en.wikipedia.org/wiki/Wikipedia:Reusing_Wikipedia_content#Re-use_of_text_under_Creative_Commons_Attribution-ShareAlike) as long as it would make the result just a modified version and not the entire library by implication a derivative work) but the algorithm originally comes from [this paper](http://www.graphics.cornell.edu/pubs/1997/MT97.pdf) I believe, and I would not know the implications of basing code on a snippet in a paper for the required license.

Uwila:
Closes #129

Tested using GCC 9.3.0 on Arch Linux. I have not tested the stuff in `win/`, as I do not program on Windows.

For some other details see [here](https://github.com/recp/cglm/issues/129#issuecomment-607814688)

### DIFF
--- a/include/cglm/ray.h
+++ b/include/cglm/ray.h
@@ -42,7 +42,7 @@ glm_ray_triangle(vec3   origin,
                  float *d) {
   vec3        edge1, edge2, p, t, q;
   float       det, inv_det, u, v, dist;
-  const float epsilon = 0.000001;
+  const float epsilon = 0.000001f;
 
   glm_vec3_sub(v1, v0, edge1);
   glm_vec3_sub(v2, v0, edge2);
@@ -54,32 +54,28 @@ glm_ray_triangle(vec3   origin,
   if (det > -epsilon && det < epsilon)
     return 0;
 
-  inv_det = 1.0 / det;
+  inv_det = 1.0f / det;
 
   glm_vec3_sub(origin, v0, t);
 
   u = inv_det * glm_vec3_dot(t, p);
 
-  if (u < 0.0 || u > 1.0)
+  if (u < 0.0f || u > 1.0f)
     return 0;
 
   glm_vec3_cross(t, edge1, q);
 
   v = inv_det * glm_vec3_dot(direction, q);
 
-  if (v < 0.0 || u + v > 1.0)
+  if (v < 0.0f || u + v > 1.0f)
     return 0;
 
   dist = inv_det * glm_vec3_dot(edge2, q);
-  if (dist > epsilon) {
-    if (d != NULL) {
+
+  if (d)
       *d = dist;
-    }
 
-    return 1;
-  }
-
-  return 0;
+  return dist > epsilon;
 }
 
 #endif
--- a/include/cglm/ray.h
+++ b/include/cglm/ray.h
@@ -7,7 +7,12 @@
 
 /*
  Functions:
-   CGLM_INLINE bool glm_line_triangle_intersect(vec3 origin, vec3 direction, vec3 v0, vec3 v1, vec3 v2, float *d);
+   CGLM_INLINE bool glm_line_triangle_intersect(vec3   origin,
+                                                vec3   direction,
+                                                vec3   v0,
+                                                vec3   v1,
+                                                vec3   v2,
+                                                float *d);
 */
 
 #ifndef cglm_ray_h
@@ -29,7 +34,12 @@
 
 CGLM_INLINE
 bool
-glm_ray_triangle(vec3 origin, vec3 direction, vec3 v0, vec3 v1, vec3 v2, float *d) {
+glm_ray_triangle(vec3   origin,
+                 vec3   direction,
+                 vec3   v0,
+                 vec3   v1,
+                 vec3   v2,
+                 float *d) {
   vec3        edge1, edge2, p, t, q;
   float       det, inv_det, u, v, dist;
   const float epsilon = 0.000001;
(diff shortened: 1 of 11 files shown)


include/cglm/ray.h
```diff
@@ -0,0 +1,66 @@
...
+ * @param[in] v0             first vertex of triangle
+ * @param[in] v1             second vertex of triangle
+ * @param[in] v2             third vertex of triangle
+ * @param[in, out] d         distance to intersection
+ * @param[out] intersection  whether there is intersection
+ */
+
+CGLM_INLINE
+bool
+glm_ray_triangle(vec3 origin, vec3 direction, vec3 v0, vec3 v1, vec3 v2, float *d) {
```
recp, line 32: You were called this functions `glm_line_triangle_intersect()`, so I guess this is **ray triangle intersection** not **line triangle intersection** right?

include/cglm/ray.h
```diff
@@ -0,0 +1,66 @@
...
+
+    glm_vec3_sub(v1, v0, edge1);
+    glm_vec3_sub(v2, v0, edge2);
+
+    glm_vec3_cross(direction, edge2, p);
+
+    det = glm_vec3_dot(edge1, p);
+
+    if (det > -epsilon && det < epsilon)
+        return 0;
```
recp, line 45: **cglm** and I use TWO SPACE for indents not TABS or 4 spaces. Also a new line after return pls.

include/cglm/ray.h
```diff
@@ -0,0 +1,66 @@
...
+
+    glm_vec3_cross(t, edge1, q);
+
+    v = inv_det * glm_vec3_dot(direction, q);
+
+    if (v < 0.0 || u + v > 1.0)
+        return 0;
+
+    *d = inv_det * glm_vec3_dot(edge2, q);
```
recp, line 62: What if user do not want to store *d* (distance)? Maybe this can be optional, in that case you should test against local variable and expose copy it to **d** if it is not null. 

I think ray tracer must be fast as possible, so adding extra *if* statement may be reconsidered.

include/cglm/ray.h
```diff
@@ -0,0 +1,66 @@
...
+ * @param[in, out] d         distance to intersection
+ * @param[out] intersection  whether there is intersection
+ */
+
+CGLM_INLINE
+bool
+glm_ray_triangle(vec3 origin, vec3 direction, vec3 v0, vec3 v1, vec3 v2, float *d) {
+    const float epsilon = 0.000001;
+    vec3 edge1, edge2, p, t, q;
+    float det, inv_det, u, v;
```
recp, line 35: I'm obsessive about code styling and sorry for that. Can you align variables pls. Maybe const assignment could be placed after all other variable definitions to make vectors be placed to the top.

include/cglm/ray.h
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright (c), Recep Aslantas.
+ *
+ * MIT License (MIT), http://opensource.org/licenses/MIT
+ * Full license can be found in the LICENSE file
+ */
+
+/*
+ Functions:
+   CGLM_INLINE bool glm_line_triangle_intersect(vec3 origin, vec3 direction, vec3 v0, vec3 v1, vec3 v2, float *d);
```
recp, line 10: I hope we are not crossing/exceeding the 80 column here. I know we are at 2020 and this limit can be changed to 120 in the future :)

recp:
One one thing, can you add (give credits) references to the algorithm and code (if there are any) to CREDITS file to respect algorithm/code owners and their works

include/cglm/ray.h
Uwila: Yep

include/cglm/ray.h
Uwila: We're at 114 here. I could add a line break somewhere if you prefer.

Uwila:
@recp I've tried implementing the changes, please have another look.

include/cglm/ray.h
```diff
@@ -0,0 +1,75 @@
...
+
+  glm_vec3_cross(t, edge1, q);
+
+  v = inv_det * glm_vec3_dot(direction, q);
+
+  if (v < 0.0 || u + v > 1.0)
+    return 0;
+
+  dist = inv_det * glm_vec3_dot(edge2, q);
+  if (dist > epsilon) {
```
recp, line 64: Maybe we don't need an if statement (extra branch I think) here? For instance:

```C

if (d != NULL) {
  *d = dist;
}

return dist > epsilon; /* single return */
```

include/cglm/ray.h
```diff
@@ -0,0 +1,75 @@
...
+  glm_vec3_sub(v2, v0, edge2);
+
+  glm_vec3_cross(direction, edge2, p);
+
+  det = glm_vec3_dot(edge1, p);
+
+  if (det > -epsilon && det < epsilon)
+    return 0;
+
+  inv_det = 1.0 / det;
```
recp, line 47: `f` suffix could be added like `1.0f` for `float` literals.

Uwila:
I've implemented the comments in your review. I'll be trying to add tests later, probably by monday/tuesday.

PR #132, opened 2020-04-06
ray_triangle tests

Uwila:
I've tried implementing tests. I've run `make check` to test if they work and that looks good to me.

I wasn't entirely sure how to write tests; I see that in lots of places you use `ASSERTIFY` in combination with some wrapper function for a specific kind of test, rather than `ASSERT`. I wasn't sure if this was just useful in those specific instances or whether I was supposed to wrap every check in another function and use `ASSERTIFY` instead. If it (or anything else) needs changing, just let me know.

### DIFF
--- a/test/src/test_ray.h
+++ b/test/src/test_ray.h
@@ -8,25 +8,25 @@
 #include "test_common.h"
 
 TEST_IMPL(GLM_PREFIX, ray_triangle) {
-  // Check whether a simple hit is recognized with the right distance
-  vec3 origin    = {0.0f, 0.0f, 0.0f};
-  vec3 direction = {1.0f, 0.0f, 0.0f};
-  vec3 opposite  = {-1.0f, 0.0f, 0.0f};
-  vec3 v0        = {5.0f, -1.0f, 1.0f};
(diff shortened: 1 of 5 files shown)


Uwila:
I just noticed I forgot to update the `vcxproj` files so I've just added a commit that does that. However, as I don't do programming on Windows these changes are untested.

recp:
Sorry I must documented how to write tests 😔

In **cglm** we use `ASSERT` to test a condition. But if you add **ASSERT**s to a util function e.g. `test_assert_mat4_eq_identity()` (*because you don't want to write loop and check if matrix is identity in multiple places*) the program won't know the existence of these ASSERTs in the util funcs.  

To fix that; util functions which use ASSERT, that returns `test_status_t` object. So the caller function (main test function) translates this to an ASSERT by wrapping that call with `ASSERTIFY` e.g. `ASSERTIFY(test_assert_mat4_eq_identity(m1))` == `ASSERT(is_identity)`.

I think it is good solution :) 

Code style again: C89 style comments `/*  */` must be used everywhere. 

Thanks for the tests 👍

I hope, we will have other ray intersect functions in the future :) And we must optimize ray functions with raw SIMD in the future to get maximum performance.

PR #130, opened 2020-03-27
Cmake: Integratrion

FMMazur:
Just some change to make possible to use with CMake projects.

Needs to implement tests.

PR #138, opened 2020-04-26
Modern CMake Integration

realaltffour:
Additional changes on top of #130: 
- Add flags for clang
- Add flags for msvc
- Use modern cmake in compiler flags handling.
- Globalize compiler options for library and tests
- Address issue with cmake output binary being bigger than make's. (Now
cmake is smaller ~176 bytes (confirmed on dynamic library on Linux)).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,14 +20,16 @@ if(CGLM_USE_C99)
   set(C_STANDARD 99)
 endif()
 
-if("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
-  list(APPEND LDFLAGS "/W4")
-elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
-  set(C_EXTENSIONS YES)
-  list(APPEND LDFLAGS "-Wall" "-Wpedantic" "-Wstrict-aliasing=2")
-  list(APPEND LDFLAGS "-Werror=strict-prototypes" "-fstrict-aliasing" "-O3")
-elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "CLANG")
-  list(APPEND LDFLAGS "-Wall" "-Weverything" "-Wpedantic")
+if(MSVC)
+	add_definitions(-DNDEBUG -D_WINDOWS -D_USRDLL -DCGLM_EXPORTS -DCGLM_DLL)
+	add_compile_options("/W3" "/Ox" "/Gy" "/Oi" "/TC" "/analyze")
+else()
+    add_compile_options("-Wall" "-Werror" "-O3")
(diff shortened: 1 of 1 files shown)


recp:
@realaltffour thank you for the contribution, I have merged the other one. Can you update your branch and resolve conflicts

CMakeLists.txt
```diff
@@ -20,14 +20,16 @@ if(CGLM_USE_C99)
...
-  list(APPEND LDFLAGS "-Werror=strict-prototypes" "-fstrict-aliasing" "-O3")
-elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "CLANG")
-  list(APPEND LDFLAGS "-Wall" "-Weverything" "-Wpedantic")
+if(MSVC)
+	add_definitions(-DNDEBUG -D_WINDOWS -D_USRDLL -DCGLM_EXPORTS -DCGLM_DLL)
+	add_compile_options("/W3" "/Ox" "/Gy" "/Oi" "/TC" "/analyze")
+else()
+    add_compile_options("-Wall" "-Werror" "-O3")
+    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+		add_compile_options("-stdlib=libc++" "-Weverything")
```
recp, line 29: > -stdlib=libc++

May I ask why do we need to add this?

CMakeLists.txt
```diff
@@ -20,14 +20,16 @@ if(CGLM_USE_C99)
...
+    else()
+		add_compile_options("-Werror=strict-prototypes" "-fstrict-aliasing")
```
recp, line 31: I think we must add `-fstrict-aliasing` to clang too?

recp:
@realaltffour thanks for your contribution 🎉

I have two questions above, hope we can get answers ;)

CMakeLists.txt
realaltffour: My bad! I didn't copy the arguments correctly! Will fix!

CMakeLists.txt
realaltffour: Yes, again, my bad!

realaltffour: @recp 
In ```clang```:
According to [this page](http://www.keil.com/support/man/docs/armclang_ref/armclang_ref_sam1465825195202.htm), the compiler automatically adds ```-fstrict-aliasing``` when the ```-O1``` or above is specified, making the specifying of the option redundant.

In ```gcc```:
According to [this page](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html), the compiler automatically adds ```-fstrict-aliasing``` when the ```-O2``` or above is specified, making the specifying of the option redundant again.

I will remove the redundant options. Thought of giving you a heads-up!

realaltffour:
@recp, new PR should be ready! #139

PR #139, opened 2020-04-26
CMake: Fix incorrect compiler flags

realaltffour:
Additional changes over #138:
Fix CMake compiler args
- Replace ```-stdlib=libc++``` with ```-std=gnu11``` & apply it to clang too.
- Remove redundant ```-fstrict-aliasing```.
- Remove warned and redundant ```-Weverything```


CMakeLists.txt
```diff
@@ -24,11 +24,12 @@ if(MSVC)
 	add_definitions(-DNDEBUG -D_WINDOWS -D_USRDLL -DCGLM_EXPORTS -DCGLM_DLL)
 	add_compile_options("/W3" "/Ox" "/Gy" "/Oi" "/TC" "/analyze")
 else()
-    add_compile_options("-Wall" "-Werror" "-O3")
+    add_compile_options("-Wall" "-Werror" "-std=gnu11" "-O3")
     if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
```
recp, line 28: Should we check compiler ID against `CMAKE_CXX_COMPILER_ID` or `CMAKE_C_COMPILER_ID ` here?

recp:
@realaltffour thanks for quick fix, just one question has left :)

CMakeLists.txt
realaltffour: Conceptually, no. Since we are doing platform checks and not version checks it is fine. `CMake` already gets the information of the `cxx` compiler even though the project is strictly `C` (check my `fixmydownloads` project's Github action's Linux logs, the project is `C` but CMake checks for cxx) NOTE: I might have missed something, and/or this doesn't apply to this projects. Adding, 'cxx' compilers does compile pure 'c' (correct me if I'm wrong), but it doesn't go the other way round)
In reality, if the project does support `cxx` (not sure here, I've read that @recp confirmed a comment on a PR that there is no need for `cxx` support, needs confirmation), then we should keep it checking `cxx`, otherwise in case of the project being strictly `C` there is no need to check `cxx` and it would be more 'correct' to check on `C`.

Also, I'd like to make it clear that we aren't actually setting the compiler, we are checking the platform.

Anyways, need @recp thoughts before making new PRs. I won't be able to make any PRs till tomorrow (it's night here🌃)

recp: @realaltffour thanks, I'm going to merge this PR, in the future we can improve this, I need do some research for CMake and its usage later.

recp:
@realaltffour thanks for your contribution 🎉

CMakeLists.txt
realaltffour: @recp. Update! CMake doesn't check `CXX` if the project is `C`. Sorry for the misinformation! The other part of the argument, whether the project supports `CXX` or doesn't, still holds true.

recp: @realaltffour I have updated that at https://github.com/recp/cglm/commit/5e6b566d6abadb2286b4905ecd5d25f40803eb09 thanks.

PR #146, opened 2020-07-02
mat3_mulv: function behaviour to match with mat4

Terri00:
Using mat3_mulv with the same source/dest vector creates undesired results. 

Respective mat4_mulv function provides temporary vec3 'res'.
Updated this function to match behaviour of mat4.

Compiling with GCC -03, the assembly code is identical to previous version when v and dest are different vectors.


recp:
Hello @Terri00 

Sorry for the late review and response, I saw the PR and forgot it for a while

Thanks for your contributions, the PR is merged now 🎉

PR #151, opened 2020-07-31
2D Affine Transforms

recp:
We have 3D transform utilities but 2D was missing. 

I added **2d** suffixes to existing functions: e.g. `glm_translate()` -> `glm_translate2d()` ... `glm_translate2d_x()`, `glm_scale2d()`...

The rotation confused me, should we only provide rotation around Z? Maybe we could provide rotate around X and Y and return new projected vector? Maybe in the future

## New functions:

**affine2d.h**:

- [x] `void glm_translate2d(mat3 m, vec2 v)`
- [x] `void glm_translate2d_to(mat3 m, vec2 v, mat3 dest)`
- [x] `void glm_translate2d_x(mat3 m, float x)`
- [x] `void glm_translate2d_y(mat3 m, float y)`
- [x] `void glm_translate2d_make(mat3 m, vec2 v)`
- [x] `void glm_scale2d_to(mat3 m, vec2 v, mat3 dest)`
- [x] `void glm_scale2d_make(mat3 m, vec2 v)`
- [x] `void glm_scale2d(mat3 m, vec2 v)`
- [x] `void glm_scale2d_uni(mat3 m, float s)`
- [x] `void glm_rotate2d_make(mat3 m, float angle)`
- [x] `void glm_rotate2d(mat3 m, float angle)`
- [x] `void glm_rotate2d_to(mat3 m, float angle, mat3 dest)`
- [x] tests
- [x] call versions
- [x] struct versions
- [x] documentation

## Improvements:

- [x] Print functions are disable unless one of these macros is defined: `DEBUG`, `_DEBUG`, `CGLM_DEFINE_PRINTS`. When running test it is nice to see print results, so `CGLM_DEFINE_PRINTS` macro is defined for test targets to print test results.


recp:
Any feedbacks would be awesome, I'll try to optimize these functions with SIMD (SSE/NEON...) after plain version is finished.

/cc @Terri00

Terri00:
Nice works, for rotation in my branch I only took the existing rotate z for 3d, and set positive angles make clockwise rotation. 

Having those extra dimensions could be cool too, ( I imagine the result being some kind of scaling effect? ). but im not sure how it would work in practice

recp:
@Terri00 thanks for your feedbacks, 

> clockwise rotation

Did you mean counter clockwise? Because with Right Hand Rule rotation must be counter clockwise?

With extra dimensions yes we will get some scale effect, maybe we must add this to TODOs and continue work on Z only. 

Another thing is that I created two functions for rotating around Z:

```C
glm_rotate2d_make(mat3 m, float angle); /* creates new rotate transform */
glm_rotate2d(mat3 m, float angle); /* rotates existing transform */
```

but should we add new function or modify existing one to store result in another matrix? This will save user to copy the result manually

```C
/* 1. */
glm_rotate2d(mat3 m, float angle, mat3 dest);

/* 2. */
glm_rotate2d_to(mat3 m, float angle, mat3 dest);
```

recp:
SIMD optimizations will be added later, struct version, tests and documentation are added, I'm going to merge this PR.

PR #154, opened 2020-08-03
win32: fix symbol exports

randy408:
Add `CGLM_STATIC` to disable the dllimport/dllexports for static builds and only export symbols for shared builds, this is needed because symbols should only appear with `dllimport` to _users_ of dynamic builds, while static builds should not define any attributes.

CMake / MSBuild hides these issues with workarounds but it causes problems with other tools. This change should be invisible to existing projects.

The rationale for replacing `CGLM_DLL` with `CGLM_EXPORTS` in `common.h` is that the latter is very close to what CMake defines automatically for shared builds (this would be `cglm_EXPORTS` in this case) and having both `CGLM_STATIC` and `CGLM_DLL` is confusing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ if(CGLM_USE_C99)
 endif()
 
 if(MSVC)
-  add_definitions(-DNDEBUG -D_WINDOWS -D_USRDLL -DCGLM_EXPORTS -DCGLM_DLL)
+  add_definitions(-DNDEBUG -D_WINDOWS -D_USRDLL)
   add_compile_options(/W3 /Ox /Gy /Oi /TC)
   
   # Ref: https://skia.googlesource.com/third_party/sdl/+/refs/heads/master/CMakeLists.txt#225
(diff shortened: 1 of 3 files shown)


randy408:
I could build with cmake + msbuild but only without tests: `LINK : fatal error LNK1104: cannot open file 'm.lib'`.

As for whether it breaks things: requiring `CGLM_DLL` outside of cmake is potentially breaking if the `cglm\call` headers are used, I could change it to require `CGLM_STATIC` as static builds didn't work until now, that means `CGLM_DLL` can be removed and the remaining macros will have the following meaning:

* `CGLM_EXPORTS` - building a shared library (dllexport)
* `CGLM_STATIC` - required for `cglm\call` headers with statically built library (no attributes)
* no defines - assume dynamic (dllimport)

randy408:
I've updated it my first comment to reflect the changes, what command/options should I use to update the VS solutions?

recp:
> CGLM_EXPORTS - building a shared library (dllexport)
CGLM_STATIC - required for cglm\call headers with statically built library (no attributes)
no defines - assume dynamic (dllimport)

liked it.

Thanks for your patience and work. It is really good to have contributors on Windows side. 

People may used `cglm/call` in their projects, we don't know. So it is better to not break their existing projects if it is possible. If we need to do that we must warn users in README and in Release Notes.

> I've updated my first comment to reflect the changes, what command/options should I use to update the VS solutions?

The output of Cmake and VS Project/Solution (in `win` folder) must be same. You can open VS Solution and make changes in project/solution settings with VS UI.

randy408:
The VS solution didn't need any changes to build, I removed the `CGLM_DLL` option to clean it up.

recp:
Thanks @randy408 , the PR is merged 🎉🤗

PR #155, opened 2020-08-03
test: don't link 'm' with msvc

randy408:
There is no separate math library for MSVC, I couldn't build the tests without this fix: `LINK : fatal error LNK1104: cannot open file 'm.lib' `.


recp:
@randy408 thanks for your contribution, the PR is merged 🎉

PR #158, opened 2020-08-04
Add meson build support

randy408:
Building the tests can be put behind an option, it just requires another `meson_options.txt` file in the top-level directory.


recp:
> Building the tests can be put behind an option

That would be awesome because it is optional in CMake and Autotools build.

recp:
@randy408 is it ready to merge?

randy408:
Yes.

recp:
@randy408 many thanks 🎉

Issue #150, opened 2020-07-26
Apple SIMD, Swift and Metal API Support and Helpers


recp:
**cglm** always tries to make things easier and faster. Currently I'm trying to switch from OpenGL to Metal and other graphics libraries by writing a GPU library (https://github.com/recp/gpu), it will be documented after first release...

**cglm** offers two APIs 

1. Array API which makes parameters easier to pass it, also easy to load CPU register
2. Struct API which provides type safety and other maybe

A few things I would like to see in **cglm**

1. Helpers to pass **cglm** types or values to Apple's platform or types.
2. Make **cglm** work better with **Swift** language, struct API will help a lot here, I think.
3. todo? 

**cglm** already provides `applesimd.h` header and it contains helpers to convert cglm types to Apple's simd types.

But there may bee another option:

---

Now let's see Apple's tempale Metal project, see `ShaderTypes.h`:

```C
typedef struct
{
    matrix_float4x4 projectionMatrix;
    matrix_float4x4 modelViewMatrix;
} Uniforms;
```

This is how I passed **cglm** values to Metal API:

```C
- (void)mtkView:(nonnull MTKView *)view drawableSizeWillChange:(CGSize)size {
  mat4      proj;
  Uniforms *uniforms;
  
  if (!(uniforms = (Uniforms*)_uniformBufferAddress))
    return;
  
  glm_perspective(glm_rad(65.0f), size.width / size.height, 0.1f, 100.0f, proj);
  
  uniforms->projectionMatrix = glm_mat4_applesimd(proj);
}
```

As you can see it contains extra copy (maybe it may be fast). We can reduce the copy with this way maybe **if this is not undefined behavior or similar**:

```C
/* common.h */
#if defined(__APPLE__)                                                        \
    && defined(SIMD_COMPILER_HAS_REQUIRED_FEATURES)                           \
    && defined(SIMD_BASE)                                                     \
    && defined(SIMD_TYPES)                                                    \
    && defined(SIMD_VECTOR_TYPES)
#  define CGLM_APPLE_SIMD
#endif
```

```C
/* types-struct.h */
typedef union CGLM_ALIGN_MAT mat4s {
  mat4  raw;
  vec4s col[4];
#if CGLM_USE_ANONYMOUS_STRUCT
  struct {
    float m00, m01, m02, m03;
    float m10, m11, m12, m13;
    float m20, m21, m22, m23;
    float m30, m31, m32, m33;
  };
#endif

/* this could be magic ??? */
[...]

PR #160, opened 2020-09-06
Added simple swiftpm package file

smumryak:
This PR adds the most simple Package.swift configuration file to enable usage of cglm in swift project.
Unfortunately swiftpm does not support header-only C targets which means this library has to either include empty dummy source for header-based version or build everything. Also using array based API in swift is not very convenient, so it makes sense to include both array based and struct based API.
The config exposes two tagets: 
- cglmc is the "all in one" cglm including inline API, both array based and struct based and the compiled files. Configuration allows usage of this target as static or dynamic library for the client
- cglm is the inline only version that includes both array based and struct based API. In order to force swiftpm to actually build this target the empty.c file is included in dummySrcForSwift folder to avoid "this target is empty" error. This target is available only as static library to give compiler and linker ability to strip unused symbols

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -15,7 +15,7 @@ let package = Package(
             path: "./",
             exclude: [
                 "./docs",
-                "./dummySrcForSwift",
+                "./src/swift",
                 "./include",
                 "./test",
                 "./win",
@@ -31,12 +31,11 @@ let package = Package(
             exclude: [
                 "./docs",
                 "./include",
-                "./src",
                 "./test",
                 "./win",
             ],
             sources: [
-                "./dummySrcForSwift",
+                "./src/swift",
             ],
             publicHeadersPath: "./include"
         ),
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,45 @@
+// swift-tools-version:5.2
+
+import PackageDescription
+
+let package = Package(
+    name: "cglm",
+    products: [
+        .library(name: "cglm", type: .static, targets: ["cglmHeader"]),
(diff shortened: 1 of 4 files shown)


recp:
@smumryak thanks for the PR.

`dummySrcForSwift` folder seems odd. 

What about creating `swift/` folder in `src/` and move empty.c to there as  `src/swift/empty.c`? In the future we can use Swift related sources in that folder if needed.

recp:
FWIW, there is a related issue about Swift and Apple SIMD library: https://github.com/recp/cglm/issues/150 any ideas, feedbacks are welcome if you have any ;) I want to make easy to work with Metal (and OpenGL, Vulkan, DirectX...) when using **cglm**...

smumryak:
hey @recp 
yea, I'll take a look if if it's possible to move that hacky empty.c inside the src directory. I'm just not sure if swift package manager will allow excluding a directory while including it's subdirectory. it's relatively young build system and a lot of stuff there might be either missing of unexpected :)
also I'm using swift under linux with vulkan, so i'm not sure i will be able to help with SIDM issues on apple's platforms

smumryak:
hey @recp, I've moved the `empty.c` file inside `src/swift` folder and made adjustments to `Package.swift`. Tried it and swiftpm successfully creates the library

smumryak:
Hmm, I think I should also update the README

recp (approved):

recp:
@smumryak the PR is merged 🎉 thanks for your contributions, I hope you keep maintain the package manager ;)

PR #164, opened 2020-11-08
simd/x86: fix -Wcast-align warnings (gcc/clang)

Akaricchi:
This modifies `glmm_{load,store}3` functions to make the compiler assume the v pointer is appropriately aligned for the type it is being cast to.

Not tested with `CGLM_ALL_UNALIGNED`, but it probably doesn't matter.

### DIFF
--- a/include/cglm/types.h
+++ b/include/cglm/types.h
@@ -32,6 +32,16 @@
 #  define CGLM_ALIGN_MAT CGLM_ALIGN(16)
 #endif
 
+#ifdef __GNUC__
+#  define CGLM_ASSUME_ALIGNED(expr, alignment) \
+  __builtin_assume_aligned((expr), (alignment))
+#else
+#  define CGLM_ASSUME_ALIGNED(expr, alignment) (expr)
+#endif
+
+#define CGLM_CASTPTR_ASSUME_ALIGNED(expr, type) \
+  ((type*)CGLM_ASSUME_ALIGNED((expr), __alignof__(type)))
+
 typedef float                   vec2[2];
 typedef float                   vec3[3];
 typedef int                    ivec3[3];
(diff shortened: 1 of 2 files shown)


include/cglm/types.h
```diff
@@ -32,6 +32,16 @@
...
 
+#ifdef __GNUC__
+#  define CGLM_ASSUME_ALIGNED(expr, alignment) \
+  __builtin_assume_aligned((expr), (alignment))
+#else
+#  define CGLM_ASSUME_ALIGNED(expr, alignment) (expr)
+#endif
+
+#define CGLM_CASTPTR_ASSUME_ALIGNED(expr, type) \
+  ((type*)CGLM_ASSUME_ALIGNED((expr), __alignof__(type)))
```
recp, line 43: I think `__alignof__` may not be standard, in this case not all compilers may support that. It would be nice to see that **cglm** works on all compilers ( including msvc ). Maybe we can check compiler support for that?

shakesoda: `alignof()` is standard since C11, not sure about `__alignof__()` or MSVC's status other than `alignof()` being supported there since C++11 (maybe they've had it in for C, too?)

Akaricchi: This doesn't matter because the second argument of the `CGLM_ASSUME_ALIGNED` macro is only expanded for the `__GNUC__` implementation. GCC (and all compilers that pretend to be GCC) is expected to have `__alignof__` in all `-std=` modes (unlike `alignof`, which may not be available in strict modes).

MSVC and the rest will get this implementation
```c
#  define CGLM_ASSUME_ALIGNED(expr, alignment) (expr)
```
Which just ignores the `alignment` argument as you can see.

recp: > alignof() is standard since C11

Good to know

> MSVC and the rest will get this implementation

Sorry, I missed the flow.

recp (approved):

recp:
@Akaricchi thanks for your contribution 🎉

PR #165, opened 2020-11-08
meson: add 'install' option

Akaricchi:
This is useful for people who want to use cglm as a meson subproject without polluting the main project's install target.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -11,6 +11,7 @@ project('cglm', 'c',
 
 cc = meson.get_compiler('c')
 
+cglm_install = get_option('install')
 cglm_deps = cc.find_library('m', required : false)
 
 cglm_args = []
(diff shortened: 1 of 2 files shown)


recp:
Hi @Akaricchi ,

There are meson build docs in `README.md` and `docs/source/build.rst`. 

Will these changes affect anything that existing doc said?:

```bash
$ meson build # [Optional] --default-library=static
$ cd build
$ ninja
$ sudo ninja install # [Optional]
```

Akaricchi:
I don't think there's anything needed to be added there, since this PR does not change the default behavior whatsoever. I guess the "Use with your Meson project" section could be expanded to show how to override the option, but that is standard Meson stuff and I think explaining it here is out of scope. I'd expect anyone trying to use this with their Meson project to have read the Meson documentation.

recp (approved):

recp:
@Akaricchi thanks for the explanation and for your contribution, the PR is merged 🎉

PR #172, opened 2020-12-02
Replace hex floating point literals

SanderMertens:


recp:
@SanderMertens thanks for the PR,

Maybe we should use a MACRO instead of that literal everywhere?

I found this comment (https://community.khronos.org/t/math-function-fract/1181/2):

> It’s a hexadecimal float constant as supported by C99, and it is indeed the largest number < 1.0 which can be represented as single precision float. It’s probably written with the 1 before the decimal point because that’s how IEEE754 floats are actually stored, with an implicit leading 1 (unless it’s a denormalized number). Then fffffe means all 23 mantissa bits are set to 1.

and this: https://www.khronos.org/registry/OpenCL/sdk/2.0/docs/man/xhtml/fract.html

also this (http://www.mitsuba-renderer.org/api/constants_8h_source.html):

```C
#if defined(__WINDOWS__)
#define ONE_MINUS_EPS_FLT 0.999999940395355225f
#define ONE_MINUS_EPS_DBL 0.999999999999999888
#define RCPOVERFLOW_FLT   2.93873587705571876e-39f
#define RCPOVERFLOW_DBL   5.56268464626800345e-309
#else
#define ONE_MINUS_EPS_FLT 0x1.fffffep-1f
#define ONE_MINUS_EPS_DBL 0x1.fffffffffffff7p-1
#define RCPOVERFLOW_FLT   0x1p-128f
#define RCPOVERFLOW_DBL   0x1p-1024
#endif
```

it would be nice to use `0x1.fffffep-1f` where it can possible and other value (` 0.999999940395355224609375f`) for everywhere else

recp:
I'll accept this PR and will update it later, thanks for your contribution 🎉

SanderMertens:
Thanks! Sorry, I didn't have the time yesterday to update it.

recp:
@SanderMertens sorry, I should have waited a bit more :) next time 🤗

Issue #181, opened 2021-03-25
Provide a CMake interface library target


legends2k:
Applications using cglm as a pure header library currently do

``` cmake
add_subdirectory(external/cglm)
target_link_libraries(MyExecutable PRIVATE cglm)
```

However, this build `.c` files under `src/` only to be thrown away since the project isn't using the compiled _call_ variant of cglm i.e. `glmc_` functions aren't used but still they're built.

Providing an interface library target and a shared library target would give freedom for users to choose the right dependency level.

**Refer**: [Creating a header-only library with CMake](http://www.mariobadr.com/creating-a-header-only-library-with-cmake.html)

recp:
@legends2k thaks for your feedbacks, you are right, we should not force users to build **cglm** if they only use headers.

Any PR would be appreciated ☺️ Otherwise I'll try to do it in my free time...

legends2k:
My understanding of CMake is slim.  Nevertheless I'll raise a PR, please do correct me if I'm doing the wrong thing.

recp:
see https://github.com/recp/cglm/pull/183

PR #183, opened 2021-03-25
Add CMake interface library target

legends2k:
Projects using cglm as a header-only library needn’t build files under
src/. Provide a target which allows them to skip compiling them by

``` cmake
add_subdirectory(external/cglm EXCLUDE_FROM_ALL)
target_link_libraries(MyExe PRIVATE cglm_headers)
```


legends2k:
Bug #181 would be resolved with this issue.

recp:
Hi @legends2k 

Thanks for creating a PR for that issue 🤗

> ```cmake
> add_subdirectory(external/cglm EXCLUDE_FROM_ALL)
> target_link_libraries(MyExe PRIVATE cglm_headers)
> ```

We should include this guide to README and docs (docs/source/build.rst) which will help to users to be aware of this feature. 

Do we need to add all headers manually in `add_target INTERFACE`?

A resource for header-only library (but we provide both compiled and header only or header-only): 
http://mariobadr.com/creating-a-header-only-library-with-cmake.html

legends2k:
@recp You're right, I've removed the header listing as it's needless.  Thanks for reviewing the change.

recp:
@legends2k the PR is merged, thanks for your contributions 🎉

PR #188, opened 2021-04-23
configure and install cglm.pc with cmake

wdouglass:
This lets non-cmake projects consume this library with pkgconfig, even if it was compiled with cmake


recp:
@wdouglass the PR is merged, thanks  🎉

PR #203, opened 2021-05-29
Update lists of source files in meson.build

bwhmather:
Adds missing files to `meson.build`, and fixes invalid syntax in entries for existing files.

Tested by running:
```
meson setup build -D build_tests=true
ninja -C build test
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -102,13 +102,17 @@ if get_option('build_tests') == true
 
 test_src = files(
     'test/runner.c',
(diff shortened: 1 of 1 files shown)


recp:
@bwhmather the PR is merged, thanks for your contribution 🎉🤗

bwhmather:
@recp - My pleasure.  Thank you for your many hundreds.  This is turning into a solid project.

PR #204, opened 2021-06-07
Include affine.h in quat.h; remove needless decls

legends2k:
Including all of `cglm/cglm.h` is sometimes an overkill.  Just including `cglm/quat.h` results in a compiler error for

> unknown symbol `glm_translate_make`

* affine.h is needed for `glm_translate_make`
* Remove function declarations
  - `glm_mat4_mulv` since `mat4.h` is already included
  - `glm_mul_rot` as `affine-mat.h` is already included
  - `glm_translate` as `affine.h` is included with this change


legends2k:
Will do a rebase pull and resubmit this PR from my fork.

PR #205, opened 2021-06-07
Include affine.h in quat.h; remove needless decls

legends2k:
Including all of `cglm/cglm.h` is sometimes an overkill.  Just including `cglm/quat.h` results in a compiler error for

> unknown symbol `glm_translate_make`

* Include `affine.h` in `quat.h` for definition of `glm_translate_make`
* Remove function declarations
  - `glm_mat4_mulv` since `mat4.h` is already included
  - `glm_mul_rot` as `affine-mat.h` is already included
  - `glm_translate` as `affine.h` is included with this change

### DIFF
--- a/include/cglm/quat.h
+++ b/include/cglm/quat.h
@@ -59,6 +59,7 @@
 #include "mat4.h"
 #include "mat3.h"
 #include "affine-mat.h"
+#include "affine.h"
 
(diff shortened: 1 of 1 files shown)


recp:
@legends2k the PR is merged, thanks for your contributions 🎉

PR #206, opened 2021-06-09
Unit quaternion vector rotation

legends2k:
This variant avoids normalize operation quat_rotatev, so it’d be
slightly faster.  The implementation is based on [Rotating a single vector using a quaternion][1].

[1]: https://fgiesen.wordpress.com/2019/02/09/rotating-a-single-vector-using-a-quaternion/


include/cglm/quat.h
```diff
@@ -745,7 +748,31 @@ glm_quat_forp(vec3 from, vec3 to, vec3 up, versor dest) {
...
- * @brief rotate vector using using quaternion
+ * @brief rotate vector using unit quaternion
+ *
+ * @param[in]   q     quaternion (must be of unit length)
+ * @param[in]   v     vector to rotate
+ * @param[out]  dest  rotated vector
+ */
+CGLM_INLINE
+void
+glm_quat_unit_rotatev(versor q, vec3 v, vec3 dest) {
```
recp, line 759: What about `glm_versor_rotatev()` as name since unit quat also kown as versors? It seems it will bring another namespace

legends2k: Yeah, `glm_versor_rotatev` is also an option but I think most games/interactive 3D apps anyways use mainly unit quaternions for representing rotation.  Is the new namespace introduction needed?

recp: >  I think most games/interactive 3D apps anyways use mainly unit quaternions for representing rotation

Hmm, I'm not sure

recp: I was used the implementation: https://gamedev.stackexchange.com/questions/28395/rotating-vector3-by-a-quaternion

I'm ot sure which one is faster, we could drop `glm_quat_normalize_to(q, p);` to expect a unit quat maybe? Did you compare speed or ASM output? Yours one contains one more glm_vec3_cross() but no glm_vec3_dot() products...

legends2k: > Yours one contains one more glm_vec3_cross() but no glm_vec3_dot() products...
Yeah, there are multiple implementations.  I saw this in the reference link shared above and also in a book.

> Did you compare speed or ASM output?

No, I didn't.  But I'm guessing a dot product is faster than a cross product, but usually it's better to measure than guess.  I know the one in PR is faster since it has no `sqrt` though.

> we could drop `glm_quat_normalize_to(q, p)` to expect a unit quat maybe?

I agree.  We can just document that the input should be a versor.  If somebody has non-normalized quaternion, they can always call it manually and then call `quat_rotatev`.

recp: Now we are modifying `quat_rotatev` to expect unit quat right? (we must add this change to README or documentation) 

Or we could duplicate `quat_rotatev`  and drop `glm_quat_normalize_to(q, p)` as `glm_quat_unit_rotatev` ? 

* `glm_vec3_rotate` expects unit **axis** , 
* `glm_quatv` normalizes the **axis** 
* `glm_rotate_make` normalizes **axis**
* `quat_rotatev`  normalizes **axis**

Making the behaviors homogeneous could be nice

legends2k: > > I think most games/interactive 3D apps anyways use mainly unit quaternions for representing rotation
> 
> Hmm, I'm not sure

Any reasons for you not begin sure?  Check https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation:

> **Unit quaternions**, known as versors, provide a convenient mathematical notation for representing spatial orientations and rotations of elements in three dimensional space.

Additional references

* https://www.gamedev.net/forums/topic/96500-non-unit-quaternions/
* https://www.gamedevs.org/uploads/understanding-quaternions.pdf (see slide 42)

So if an application is using quaternions for rotation, they'd be of unit length.

legends2k: For the homogeneity part, actually all four functions normalize the passed axis; except `glm_vec3_rotate` other functions have the correct documentation.  `glm_vec3_rotate`'s documentation says  _must be unit vector_ but implementation calls `glm_vec3_normalize_to(axis, k)` 🤕

Based on the above rationale of giving the application the choice to normalize or skip, I'd suggest we should simply expect unit axis in all 4 functions.  Please let me know what you think.

recp: > Just expecting unit quaternions and leaving option of normalizing to the application seems like the optimal choice to me. This doesn't waste cycles needlessly.

I like your approach, let's make expectations about axes: always unit. 

> glm_vec3_rotate's documentation says must be unit vector but implementation calls glm_vec3_normalize_to(axis, k) 🤕

😲 🤦‍♂️

> Based on the above rationale of giving the application the choice to normalize or skip, I'd suggest we should simply expect unit axis in all 4 functions and drop the normalize_to calls. Please let me know what you think.

Let's do it.

recp: With this decision, there is no need for `glm_quat_unit_rotatev` anymore.

legends2k:
Based on the above discussions closing this PR without merging to raise a separate PR for making all axes/quaternions unit vector.

PR #207, opened 2021-06-09
Fix vec3_ortho

legends2k:
* Fix `glm_vec3_ortho` to work with all `vec3` inputs
  - Previous algorithm results in 0-vector for inputs like `<1, 1, 1>`
* Update documentation with algorithm reference link


recp:
Hi @legends2k good catch and thanks for fixing this

recp:
@legends2k many thanks, the PR is merged 🎉

Issue #208, opened 2021-06-10
`glm_vec3_cross` doesn't work when `dest` is one of the operands


legends2k:
`glm_vec3_cross` implementation implicitly assumes that all three arguments `a`, `b` and `dest` are different.  Code like

```c
vec3 a = {0.f, 1.f, 0.f}, b = {1.f, 0.f, 0.f};
glm_vec3_cross(a, b, a);  // oops! a is now a zero vector, not -Z as expected
```

where `a` and `dest` are the same doesn't work due to this.

recp:
```C
CGLM_INLINE
void
glm_vec3_cross(vec3 a, vec3 b, vec3 dest) {
  /* (u2.v3 - u3.v2, u3.v1 - u1.v3, u1.v2 - u2.v1) */
  dest[0] = a[1] * b[2] - a[2] * b[1];
  dest[1] = a[2] * b[0] - a[0] * b[2];
  dest[2] = a[0] * b[1] - a[1] * b[0];
}
```

You are right, we should duplicate one of vectors or both to avoid this.

recp:
@legends2k many thanks for the catch and fix.

Related PR: https://github.com/recp/cglm/pull/209

PR #209, opened 2021-06-10
Fix cross product when an operand is also dest

legends2k:
Fix cross product function to work when `dest` is one of the operands by
storing a result in a temporary and then copying it to `dest`.

Fix for #208


recp:
@legends2k many thanks, the PR is merged 🎉🚀

PR #210, opened 2021-06-10
Compute quaternion rotating a vector into another

legends2k:
* Given two unit `vec3`s, compute versor that would rotate `a` into `b`
* Add tests for verify following cases
  - Parallel
  - Perpendicular
  - Straight
  - Random
  - Acute
  - Obtuse
  - Zero

### DIFF
--- a/include/cglm/quat.h
+++ b/include/cglm/quat.h
@@ -197,22 +197,23 @@ glm_quat_copy(versor q, versor dest) {
 CGLM_INLINE
 void
 glm_quat_from_vecs(vec3 a, vec3 b, versor dest) {
-  float cos_theta = glm_vec3_dot(a, b);
-  if (cos_theta >= 1.f - GLM_FLT_EPSILON) {  // a ∥ b
+  CGLM_ALIGN(8) vec3 axis;
+  float cos_theta;
+  float cos_half_theta;
+
+  cos_theta = glm_vec3_dot(a, b);
+  if (cos_theta >= 1.f - GLM_FLT_EPSILON) {  /*  a ∥ b  */
     glm_quat_identity(dest);
     return;
   }
-  CGLM_ALIGN(8) vec3 axis;
-  float cos_half_theta;
-  if (cos_theta < -1.f + GLM_FLT_EPSILON) {  // angle(a, b) = 180°
+  if (cos_theta < -1.f + GLM_FLT_EPSILON) {  /*  angle(a, b) = π  */
     glm_vec3_ortho(a, axis);
-    cos_half_theta = 0.f;
-  }
-  else {
+    cos_half_theta = 0.f;                    /*  cos π/2 */
+  } else {
     glm_vec3_cross(a, b, axis);
-    const float cos_zero = 1.0f;
-    cos_half_theta = cos_zero + cos_theta;
+    cos_half_theta = 1.0f + cos_theta;       /*  cos 0 + cos θ  */
   }
+
   glm_quat_init(dest, axis[0], axis[1], axis[2], cos_half_theta);
   glm_quat_normalize(dest);
 }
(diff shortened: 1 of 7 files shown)


legends2k:
@recp Note: this PR (creating a quaternion from two vectors) is unrelated and different from [the previous one][1] (rotating a vector using quaternion).

Once this PR is in, I'll create another PR for resolving the unit axes issue.  It'll make all functions taking an input axes into unit axes and also update documentation.

[1]: https://github.com/recp/cglm/pull/206

recp:
@legends2k sorry for the late response,

We already have `glm_quat_for()` and `glm_quat_for()`, are there any differences with them?

legends2k:
> We already have `glm_quat_for()` and `glm_quat_forp()`, are there any differences with them?

Existing `glm_quat_for` is used to compute an [orthonormal basis][] for entities like camera.  The rotation needed to go from current space to such space would be encoded in the destination quaternion.  It takes forward and up vectors as inputs, generates a side vector and forms such a basis.

This function is different; it just gives the rotation between two vectors.  Say a model is looking towards <0.707, 0.707, 0> and we want it to look towards <0, 1, 0>, feeding these two vectors would give the needed rotation as a quaternion.  The existing function would not do this; it's simply for a different purpose.

[orthonormal basis]: https://en.wikipedia.org/wiki/Orthonormal_basis

recp:
@legends2k great! 

Can you declare all variables at the beginning of func/block like (https://github.com/recp/cglm/blob/master/CONTRIBUTING.md):

```C
void
func() {
  float variable1, variable2;
  int   variable3;

  variable1 = 123.0f;
}
```

include/cglm/quat.h
```diff
@@ -184,10 +187,40 @@ glm_quat_copy(versor q, versor dest) {
...
+  }
+  CGLM_ALIGN(8) vec3 axis;
+  float cos_half_theta;
+  if (cos_theta < -1.f + GLM_FLT_EPSILON) {  // angle(a, b) = 180°
+    glm_vec3_ortho(a, axis);
+    cos_half_theta = 0.f;
+  }
+  else {
+    glm_vec3_cross(a, b, axis);
+    const float cos_zero = 1.0f;
```
recp, line 213: I think there is no need to create separate consst variable. Float literal woul also be a constant: `1.0f + cos_theta`

include/cglm/quat.h
```diff
@@ -184,10 +187,40 @@ glm_quat_copy(versor q, versor dest) {
...
+    glm_quat_identity(dest);
+    return;
+  }
+  CGLM_ALIGN(8) vec3 axis;
+  float cos_half_theta;
+  if (cos_theta < -1.f + GLM_FLT_EPSILON) {  // angle(a, b) = 180°
+    glm_vec3_ortho(a, axis);
+    cos_half_theta = 0.f;
+  }
+  else {
```
recp, line 211: No new-line for else pls.

include/cglm/quat.h
```diff
@@ -184,10 +187,40 @@ glm_quat_copy(versor q, versor dest) {
...
+void
+glm_quat_from_vecs(vec3 a, vec3 b, versor dest) {
+  float cos_theta = glm_vec3_dot(a, b);
+  if (cos_theta >= 1.f - GLM_FLT_EPSILON) {  // a ∥ b
+    glm_quat_identity(dest);
+    return;
+  }
+  CGLM_ALIGN(8) vec3 axis;
+  float cos_half_theta;
+  if (cos_theta < -1.f + GLM_FLT_EPSILON) {  // angle(a, b) = 180°
```
recp, line 207: C89 style comment pls e.g `/*  */`

recp (changes requested):

include/cglm/quat.h
legends2k: Done!

legends2k:
> @legends2k great!
> 
> Can you declare all variables at the beginning of func/block like (https://github.com/recp/cglm/blob/master/CONTRIBUTING.md):

Done!

include/cglm/quat.h
legends2k: Done!

legends2k:
@recp Incorporated PR comments, please check now :)

recp:
@legends2k many thanks 👏 the pr is merged

PR #214, opened 2021-08-09
Note that radians is expected in glm_perspective()

bubba2k:
The `fovy` parameter is expected in radians, which isn't mentioned anywhere in the docs. Causes unnecessary confusion. Might as well point it out.


recp:
Hi @bubba2k 

Thanks for your contribution, the PR is merged 🚀

FWIW, all angles are radians in **cglm**

PR #217, opened 2021-08-15
Add CGLM_FORCE_DEPTH_ZERO_TO_ONE support to glm_unproject

### LINKED ISSUE #216: glm_unproject not working with CGLM_FORCE_DEPTH_ZERO_TO_ONE
I recently made a switch to using zero-to-one for my depth buffer (OpenGL 4.5). I have `CGLM_FORCE_DEPTH_ZERO_TO_ONE` defined, and have a call to `glClipControl( GL_LOWER_LEFT, GL_ZERO_TO_ONE )`.

However, with these changes, I noticed that `glm_unproject()` started giving weird results. I use it in my game for placing 3D objects at the mouse cursor, and they no longer line up properly. If I remove the `glClipControl` call, things appear as they should.

I'm guessing `glm_unproject()` doesn't use the `CGLM_FORCE_DEPTH_ZERO_TO_ONE` setting yet?

Nairou:
I don't know if this is the style you want to use for toggling on the flag, but the change works in my testing.

Fixes #216

### DIFF
--- a/include/cglm/project.h
+++ b/include/cglm/project.h
@@ -46,7 +46,11 @@ glm_unprojecti(vec3 pos, mat4 invMat, vec4 vp, vec3 dest) {
 
   v[0] = 2.0f * (pos[0] - vp[0]) / vp[2] - 1.0f;
   v[1] = 2.0f * (pos[1] - vp[1]) / vp[3] - 1.0f;
+#ifdef CGLM_FORCE_DEPTH_ZERO_TO_ONE
+  v[2] = pos[2];
+#else
   v[2] = 2.0f *  pos[2]                  - 1.0f;
+#endif
   v[3] = 1.0f;
 
   glm_mat4_mulv(invMat, v, v);


recp:
@Nairou many thanks for the PR,

Great! But I think it would be awesome if we follow same way as existing api. For instance we can move clipspace related implementations into `include/cglm/clipspace` folder and use macros to include them. You can see the implementation in `cglm/cam.h` header.

If you don't have enough time (I wish you have ☺️) I'll accept the PR and expect new PRs to improve it or I'll try to do it myself

recp:
GLM implementation:

https://github.com/g-truc/glm/blob/b3f87720261d623986f164b2a7f6a0a938430271/glm/ext/matrix_projection.hpp
https://github.com/g-truc/glm/blob/b3f87720261d623986f164b2a7f6a0a938430271/glm/ext/matrix_projection.inl

It is `ZO` (ZERO-to-ONE) and `NO` (Negative-One), it seems no need to implement `LH` versions

Nairou:
I'll see what I can do. Do you want `LH` copies as duplicates of `RH`, or just have the `LH` flags include the `RH` versions?

Nairou:
I'll open a new PR as this one was a simple Github edit

recp:
Thanks, 

> Do you want LH copies as duplicates of RH, or just have the LH flags include the RH versions?

According to GLM implementation, it seems thereare no LH/RH implementations for project/unproject. So we can follow it for now. We should also investigate that if we need to LH/RH separation if there are any difference... 

it could be like (assuming no need to LH/RH separation)

```C
CGLM_INLINE
void
glm_unproject(vec3 pos, mat4 m, vec4 vp, vec3 dest) {
#if CGLM_CONFIG_CLIP_CONTROL & CGLM_CLIP_CONTROL_ZO_BIT
  glm_unproject_zo(eye, dir, up, dest);
#elif CGLM_CONFIG_CLIP_CONTROL & CGLM_CLIP_CONTROL_NO_BIT
  glm_unproject_no(eye, dir, up, dest);
#endif
}
```

in this case, in the `clipspace` folder, we would have only `project_zo.h` and `project_no.h` headers...

Issue #227, opened 2021-11-06
Add explicit complex type or complex number support in `vec2-ext`


mxi:
I implemented the Newton-Raphson root finder visualizer with the help of cglm for some view transformations. To my surprise during development, I wasn't able to find any complex number routines in cglm and had to write a poor man's implementation. 

I imagine the argument would be to simply use versors with the `j` and `k` components set to zero, but it's a bit messy jumping between `mat4`/`vec3` and `mat2`/`mat3`/`vec2`.

Would there be any objections to implementing some routines like complex multiplication, division, etc. in `vec2-ext`? I can take that on given a few days to adjust to the codebase a bit more, but I want to make sure my work isn't wasted.

Related: #173

Thanks,
Average cglm enjoyer

recp:
Hi @mxi , sorry to the delay,

We can consider to support complex numbers if it is worth it,

> I can take that on given a few days to adjust to the codebase a bit more, but I want to make sure my work isn't wasted.

Did you implement complex numbers in your local branch?

Thanks

mxi:
> Did you implement complex numbers in your local branch?

Did just now (#229).

recp:
That PR is merged, thanks 🚀

PR #229, opened 2021-11-09
Add complex routines to `vec2-ext`

mxi:
Related discussion: #227 

In the end only 3 new functions are introduced for multiplication, division, and conjugation—hence why I don't think a separate `complex` type is necessary.

I've tried my best to add tests, and while they seem to work, I had to hack it together a bit since I didn't see any other `vec2-ext` functions being tested and I had to do improvisations of sorts.


recp:
@mxi the PR is merged, thanks for your contributionn 🚀

PR #231, opened 2021-11-21
set PACKAGE_VERSION so it gets substituted in cglm.pc.in by cmake

omar-polo:
Hello,

I noticed that the installed `cglm.pc.in` file generated by cmake ends up with a bogus line `Version:`: the `PACKAGE_VERSION` is empty!  It was probably forgot in #188.

With this patch applied, the pc file has the correct Version line.


recp:
@omar-polo thanks for your contribution, the PR is merged 🚀

omar-polo:
thanks ;)

PR #243, opened 2022-03-08
Fix a bug in glm_lookat_lh

Sand3r-:
I've noticed that the `glm_lookat_lh` produces incorrect results (X axis being negated) which cause the winding order of the triangles to be inversed. By looking into the code, I've noticed that in the normalized cross product between `up` and `f`, as well as the cross product of `f` and `s` have their arguments in the opposite order.

Both glm:
https://github.com/g-truc/glm/blob/b3f87720261d623986f164b2a7f6a0a938430271/glm/ext/matrix_transform.inl#L125-L126
And DirectXMath:
https://github.com/microsoft/DirectXMath/blob/main/Inc/DirectXMathMatrix.inl#L2221-L2224
Show the correct way to perform these cross products in a Left-Hand Coordinate system.

### DIFF
--- a/include/cglm/clipspace/view_lh.h
+++ b/include/cglm/clipspace/view_lh.h
@@ -38,7 +38,7 @@ glm_lookat_lh(vec3 eye, vec3 center, vec3 up, mat4 dest) {
   glm_vec3_normalize(f);
 
   glm_vec3_crossn(up, f, s);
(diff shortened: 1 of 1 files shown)


recp:
@Sand3r- many thanks for fixing this 🤗 the PR is merged now 🚀 Feel free to fix anything else or report as an issue

PR #244, opened 2022-03-18
Remove unused variable

geckor14:
The most recent version of Apple Clang (13.1.6) on macOS warns r3 is an unused variable. Since -Werror is enabled, this breaks the build.


recp:
@geckor14 many thanks for your contribution, the PR is merged 🚀

PR #250, opened 2022-05-19
Fix pkg-config generation when CMAKE_INSTALL_*DIR is absolute

### LINKED ISSUE #249: pkg-config file is broken when CMAKE_INSTALL_{INCLUDE,LIB}DIR is absolute
As per title: `CMakeLists.txt` has
```
set(prefix ${CMAKE_INSTALL_PREFIX})
set(exec_prefix ${CMAKE_INSTALL_PREFIX})
set(includedir "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
set(libdir "\${prefix}/${CMAKE_INSTALL_LIBDIR}")
```
and so can’t handle absolute paths in `CMAKE_INSTALL_{INCLUDE,LIB}DIR`. This leads to broken .pc files on NixOS in particular. (Using `prefix` and not `exec_prefix` to derive `libdir` also seems a bit suspect, but is not the problem here.)

See “[Concatenating paths when building pkg-config files](https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files)” for a thorough discussion of the problem and a suggested fix, or [KDE’s extra-cmake-modules](https://invent.kde.org/frameworks/extra-cmake-modules/-/blob/master/modules/ECMGeneratePkgConfigFile.cmake#L166) for a simpler approach.

alexshpilkin:
Fixes #249.


recp:
@alexshpilkin the PR is merged, thanks 🚀

PR #254, opened 2022-07-15
Update README to show that integer types are supported

Chris-F5:


recp:
@Chris-F5 thanks for the fix, the PR is merged 🚀

PR #260, opened 2022-10-18
new affine transform functions 🚀

recp:
`glm_spin()` is good one I think. `glm_rotate()` rotates around origin it would be enough to use `glm_rotate_at()` but `glm_spin()` is better name if you want to rotate around itself, not origin nor around any specific point.

### new functions:
---

- [x] **`CGLM_INLINE void glm_spin(mat4 m, float angle, vec3 axis)`**: rotates existing transform around itself by with axis. this helps to rotate object around itself without moving manually or providing a pivot (***which will be position of object anyway for this case***) for `glm_rotate_at()` to rotate around itself 

---

Current implementations apply transforms to vector first, but sometimes we may need to apply rotations as last transform, **cglm** introduces post-transform operations e.g. apply rotation/translate last instead of first to vector:

- [x]   `CGLM_INLINE void glm_translated_to(mat4 m, vec3 v, mat4 dest)`
- [x]   `CGLM_INLINE void glm_translated(mat4 m, vec3 v)`
- [x]   `CGLM_INLINE void glm_translated_x(mat4 m, float to)`
- [x]   `CGLM_INLINE void glm_translated_y(mat4 m, float to)`
- [x]   `CGLM_INLINE void glm_translated_z(mat4 m, float to)`
- [x]   `CGLM_INLINE void glm_rotated_x(mat4 m, float angle, mat4 dest)`
- [x]   `CGLM_INLINE void glm_rotated_y(mat4 m, float angle, mat4 dest)`
- [x]   `CGLM_INLINE void glm_rotated_z(mat4 m, float angle, mat4 dest)`
- [x]   `CGLM_INLINE void glm_rotated(mat4 m, float angle, vec3 axis)`
- [x]   `CGLM_INLINE void glm_rotated_at(mat4 m, vec3 pivot, float angle, vec3 axis)`
- [x]   `CGLM_INLINE void glm_spinned(mat4 m, float angle, vec3 axis)`

---

new project functions:

- [x]   `CGLM_INLINE float glm_project_z(vec3 pos, mat4 m)` - map object's z coordinate to window coordinates
- [x]   `CGLM_INLINE float glm_project_z_no(vec3 pos, mat4 m)`
- [x]   `CGLM_INLINE float glm_project_z_zo(vec3 pos, mat4 m)`

names are similar to existing ones except I added `ed` suffix to existing functions but we can use better names if there are any suggestions?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -159,6 +159,8 @@ cglm_structdir=$(includedir)/cglm/struct
 cglm_struct_HEADERS = include/cglm/struct/mat4.h \
                       include/cglm/struct/mat3.h \
                       include/cglm/struct/mat2.h \
+                      include/cglm/struct/affine-pre.h \
+                      include/cglm/struct/affine-post.h \
                       include/cglm/struct/affine.h \
                       include/cglm/struct/affine2d.h \
                       include/cglm/struct/vec2.h \
--- a/docs/source/affine-common.rst
+++ b/docs/source/affine-common.rst
@@ -1,0 +1,129 @@
+.. default-domain:: C
+
+3D Affine Transforms (common)
+================================================================================
+
+Common transfrom functions.
+
+Table of contents (click to go):
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Functions:
+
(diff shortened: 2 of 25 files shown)


recp:
instead of `glm_rotated()` / `glm_translated()` we could use better names maybe like `glm_addrotate()`, `glm_addtranslate()`, `glm_posttranslate()`, `glm_postrotate()`... but first one is more short version which I like :)

PR #279, opened 2023-03-06
Check for builtin before using it

NeRdTheNed:
This fixes building cglm with compilers that define __GNUC__ and don't implement __builtin_assume_aligned.

Note that the check for __has_builtin being defined and using the __has_builtin() macro need to be on different lines, as when __has_builtin is not defined, using the __has_builtin() macro is an invalid preprocessor directive.


recp:
Hi @NeRdTheNed, the PR is merged, many thanks for your contributions 🚀

Issue #278, opened 2023-02-26
NEON literal assignment


gottfriedleibniz:
Hi,

There are a few instances of direct literal assignment of float32x4_t's that will not compile w/ MSVCs Arm64 NEON declarations (does not support GCC/clang-style vector extensions; reference [this example](https://godbolt.org/z/rnb87GT1c)):

- [mat4.h#L111](https://github.com/recp/cglm/tree/bc8dc72739e4adec2a730792800dc82a2ff9f17d/include/cglm/simd/neon/mat4.h#L111)
- [mat4.h#L184](https://github.com/recp/cglm/tree/bc8dc72739e4adec2a730792800dc82a2ff9f17d/include/cglm/simd/neon/mat4.h#L184)
- [quat.h#L26](https://github.com/recp/cglm/tree/bc8dc72739e4adec2a730792800dc82a2ff9f17d/include/cglm/simd/neon/quat.h#L26)

While, MSC does not define [\_\_ARM_NEON](https://github.com/recp/cglm/blob/master/include/cglm/simd/intrin.h#L66-L68) nor __ARM_NEON_FP by default. The above statements seem to be the only thing preventing its use under MSVC Arm64.

recp:
Hi @gottfriedleibniz,

Sorry for the delay, I just saw the issue. 

> MSC does not define [__ARM_NEON](https://github.com/recp/cglm/blob/master/include/cglm/simd/intrin.h#L66-L68) nor __ARM_NEON_FP by default

Wow, I'll investigate NEON on MSVC asap, any suggestions?

recp:
@gottfriedleibniz can you check (and test if possible) https://github.com/recp/cglm/issues/278 PR. Any feedbacks would help a lot, thanks

recp:
It uses `glmm_float32x4_init()` macro to init float32x4 on neon. 

Definition: 

```C
#if defined(_WIN32) && defined(_MSC_VER)
#  define glmm_float32x4_init(x, y, z, w) { .n128_f32 = { x, y, z, w } }
#else
#  define glmm_float32x4_init(x, y, z, w) { x, y, z, w }
#endif
```

and usage:

```C
/* glm_mat4_det_neon() line 111: */
float32x4_t   x3 = glmm_float32x4_init(0.f, -0.f, 0.f, -0.f);
```

PR #280, opened 2023-03-07
fix building ARM NEON on windows & msvc

recp:
Some improvements & fixes on using ARM NEON on Windows & MSVC. Fixes https://github.com/recp/cglm/issues/278

__ARM_NEON is defined  as:
```C
/* ... */
#  elif defined(_M_ARM64)
#  ifndef __ARM_NEON
#    define __ARM_NEON
#    define __ARM_NEON_FP
#  endif
```

maybe we should check more to define it?

---

It uses `glmm_float32x4_init()` macro to init float32x4 on NEON then:

Definition: 

```C
#if defined(_WIN32) && defined(_MSC_VER)
#  define glmm_float32x4_init(x, y, z, w) { .n128_f32 = { x, y, z, w } }
#else
#  define glmm_float32x4_init(x, y, z, w) { x, y, z, w }
#endif
```

and usage:

```C
/* glm_mat4_det_neon() line 111: */
float32x4_t   x3 = glmm_float32x4_init(0.f, -0.f, 0.f, -0.f);
```


gottfriedleibniz:
Some (unorganized) thoughts:

1. Given that `CGLM_ARM64` is [defined](https://github.com/recp/cglm/blob/master/include/cglm/simd/arm.h#L14) elsewhere. This PR could also include a check for `_M_ARM`.
1. To match the ARM [documentation](https://developer.arm.com/documentation/102467/0100/Program-conventions), `__ARM_NEON` to `__ARM_NEON_FP` could be set to `1`. 
1. The use of designated initializers may not play nicely w/ C++ (or ANSI C), see this [example](https://godbolt.org/z/sGMrjfdGh). To get this example to compile, you need to include `/TC`.

The example above also includes alternative implementations. Although, I would like to see how other libraries handle this.

recp:
@gottfriedleibniz many thanks for reviewing the PR and your feedbacks. The commit (https://github.com/recp/cglm/pull/280/commits/a30baffafa0b5f0536098bedbe3896f434706512) addressed your some feedbacks I guess.  Currently all ARM stuff are related to NEON so I postponed **_M_ARM** for now, I'm not sure how it could help now (maybe later?)

> The use of designated initializers may not play nicely w/ C++ (or ANSI C), see this [example](https://godbolt.org/z/sGMrjfdGh). To get this example to compile, you need to include /TC.

You are right but is there another way to make NEON run on MSVC? If it is necessary we can disable NEON on ANSI C if it cannot compile designated initializers?

> The example above also includes alternative implementations. Although, I would like to see how other libraries handle this.

Sure I'll try to find some libraries on Github to see what is done

gottfriedleibniz:
From my example link given above, using `glmm_set4_intrin` as an inline function is probably the simplest solution (replacing `__declspec(align...` with `CGLM_ALIGN`).  

However, the current `CGLM_ARM64` check is slightly wrong (it will also be enabled on A32+NEON).

recp:
> From my example link given above, using glmm_set4_intrin as an inline function (CGLM_INLINE) is probably the simplest solution (also replacing __declspec(align... with CGLM_ALIGN).

@gottfriedleibniz thanks I updated the PR

> However, the current CGLM_ARM64 check in this PR is slightly wrong: it will also be enabled on A32+NEON.

What changes do you suggest to make it right then?

gottfriedleibniz:
> What changes do you suggest to make it right then?

With the current changes, maybe something like:

```c
#if defined(__ARM_NEON) || defined(__ARM_NEON__)
#  include <arm_neon.h>
#  define CGLM_NEON_FP 1
#  ifndef CGLM_SIMD_ARM
#    define CGLM_SIMD_ARM
#  endif
#endif
```

This matches how AVX/SSE4/SS2/etc. are handled.

1. The [previous](https://github.com/recp/cglm/pull/280/files#diff-722fe04f38f242abc9a02bb048684fcb33a8cb326a491241847712fccf44e4d4L13-L16) `CGLM_ARM64` check is fine, although, maybe inserting an `#else\n# define CGLM_ARM64 0` to quash macro undefined warnings on A32 would be nice.
1. `__ARM_NEON_FP` can be removed as it is unused.
1. MSVC solutions (or build scripts) will just have to include `/D__ARM_NEON` if their toolchain doesn't include the definition (`__ARM_NEON__` is mostly a legacy define at this point).

recp:
Can you check again with last changes? 

```C
/* ARM Neon */
#if defined(_WIN32)
/* TODO: non-ARM stuff already inported, will this be better option */
/* #  include <intrin.h> */

#  if defined(_M_ARM64) || defined(_M_HYBRID_X86_ARM64) || defined(_M_ARM64EC)
#    include <arm64intr.h>
#    include <arm64_neon.h>
#    ifndef CGLM_NEON_FP
#      define CGLM_NEON_FP  1
#    endif
#    ifndef CGLM_SIMD_ARM
#      define CGLM_SIMD_ARM
#    endif
#  elif defined(_M_ARM)
#    include <armintr.h>
#    include <arm_neon.h>
#    if defined(CGLM_NEON_FP) && (defined(__ARM_NEON_FP) || defined(vaddq_f32)) /* vaddq_f32 is defined as macro, we pick it */
#      define CGLM_NEON_FP 1
#    endif
#    ifndef CGLM_SIMD_ARM
#      define CGLM_SIMD_ARM
#    endif
#  endif

#else /* non-windows */
#  if defined(__ARM_NEON) || defined(__ARM_NEON__)
#    include <arm_neon.h>
#    if defined(__ARM_NEON_FP)
#      define CGLM_NEON_FP 1
#    endif
#    ifndef CGLM_SIMD_ARM
#      define CGLM_SIMD_ARM
#    endif
#  endif
#endif
```

There is no such NEON related macros at: https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

So I checked an existing macro to check existence of NEON on A32. 

Another fix:

```C
#if defined(_M_ARM64) || defined(_M_HYBRID_X86_ARM64) || defined(_M_ARM64EC) || defined(__aarch64__)
# define CGLM_ARM64 1
#else
# define CGLM_ARM64 0
#endif
```

References:
https://github.com/embree/embree/blob/master/common/sys/intrinsics.h

gottfriedleibniz:
I think it would be easier to keep this simple and not fight against MSVC trying to be different. For example, referencing an earlier [commit](https://github.com/recp/cglm/pull/280/commits/a30baffafa0b5f0536098bedbe3896f434706512): moving the `CGLM_ARM64` definition to `intrin.h` and using it: `#if defined(__ARM_NEON) || defined(__ARM_NEON__) || CGLM_ARM64` covers most use-cases.

Also, some uses of `CGLM_NEON_FP` over `CGLM_SIMD_ARM` seem inconsistent, e.g.,

```c
#if defined(CGLM_SIMD_x86) || defined(CGLM_NEON_FP) /* Why not CGLM_SIMD_ARM? */
```

Other references:
1. [glm](https://github.com/g-truc/glm/blob/master/glm/simd/platform.h#L333-L336) checks `__ARM_ARCH`.
1. [Eigen](https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Core/util/ConfigureVectorization.h#L377-381) checks `__ARM_NEON`.
1. [DirectXMath](https://github.com/microsoft/DirectXMath/blob/main/Inc/DirectXMath.h#L85) assumes NEON by default.

recp:
I updated a bit and it seems good to me.

> #if defined(__ARM_NEON) || defined(__ARM_NEON__) || CGLM_ARM64

There is a case where ARM is supported and ARM64 is not. The last changes handles both

recp:
@gottfriedleibniz the PR is merged now, feel free to create new one for new improvements 🚀

PR #291, opened 2023-04-21
simd, sse: use 0x80000000 insteaf of -0.f to fix fastmath on msvc

recp:
fixes https://github.com/recp/cglm/issues/289

Now we have global macro for sign masks:

```C

#define GLMM_NEGZEROf 0x80000000 /*  0x80000000 ---> -0.0f  */

#define GLMM__SIGNMASKf(X, Y, Z, W)                                           \
   _mm_castsi128_ps(_mm_set_epi32(X, Y, Z, W))
  /* _mm_set_ps(X, Y, Z, W); */

#define glmm_float32x4_SIGNMASK_PNPN GLMM__SIGNMASKf(0, GLMM_NEGZEROf, 0, GLMM_NEGZEROf)
#define glmm_float32x4_SIGNMASK_NPNP GLMM__SIGNMASKf(GLMM_NEGZEROf, 0, GLMM_NEGZEROf, 0)
#define glmm_float32x4_SIGNMASK_NPPN GLMM__SIGNMASKf(GLMM_NEGZEROf, 0, 0, GLMM_NEGZEROf)

#define glmm_float32x4_SIGNMASK_NEG _mm_set1_ps(GLMM_NEGZEROf)
```

for instance:
 `_mm_set_ps(-0.f, 0.f, -0.f, 0.f)` -> `glmm_float32x4_SIGNMASK_NPNP `
`_mm_set_ps(0.f, -0.f, 0.f, -0.f)` -> `glmm_float32x4_SIGNMASK_PNPN `

PR #292, opened 2023-04-22
wasm: replace usage of -0.f to 0x80000000

myfreeer:
Part of https://github.com/recp/cglm/pull/291

Should note that only `glmm_float32x4_SIGNMASK_PNPN` is used in wasm simd128.

This should not change the compiled output, according to [compiler explorer](https://godbolt.org/z/KjKfWzTWG)
compiler explorer

### DIFF
--- a/include/cglm/simd/wasm.h
+++ b/include/cglm/simd/wasm.h
@@ -26,6 +26,16 @@
 #define glmm_splat_z(x) glmm_splat(x, 2)
 #define glmm_splat_w(x) glmm_splat(x, 3)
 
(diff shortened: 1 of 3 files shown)


recp:
@myfreeer many thanks for fast response and for the PR. The PR is merged 🚀

myfreeer:
@recp It seems you have merged this after https://github.com/recp/cglm/pull/291, this pr is targeting [recp:fix-simd-signmask](https://github.com/recp/cglm/tree/fix-simd-signmask) to be part of https://github.com/recp/cglm/pull/291, so this is not yet part of master.

recp:
@myfreeer my mistake sorry, https://github.com/recp/cglm/pull/293 brings the changes to master.

PR #297, opened 2023-05-06
add glms_translate_to() 

kagetu:
add glms_translate_to() just like others.

### DIFF
--- a/include/cglm/struct/affine-pre.h
+++ b/include/cglm/struct/affine-pre.h
@@ -7,6 +7,7 @@
 
 /*
  Functions:
+   CGLM_INLINE mat4s glms_translate_to(mat4s m, vec3s v, mat4s dest);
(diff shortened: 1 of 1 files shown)


recp:
Hi @kagetu,

Actually struct api doesn't need to `_to` versions since functions already return result as `dest`


```C
mat4s result = glms_translate(matrix, ...)
```

since the matrix is value type result is copy of the translated **matrix** ( or should be )

kagetu:
okay, i mistake, nvm

> Hi @kagetu,
> 
> Actually struct api doesn't need to `_to` versions since functions already return result as `dest`
> 
> ```c
> mat4s result = glms_translate(matrix, ...)
> ```
> 
> since the matrix is value type result is copy of the translated **matrix** ( or should be )

PR #299, opened 2023-05-13
mat4: add new function glm_mat4_make

EasyIP2023:
Let me know if I'm missing anything please.

### DIFF
--- a/include/cglm/struct/mat4.h
+++ b/include/cglm/struct/mat4.h
@@ -459,4 +459,17 @@ glms_mat4_(rmc)(vec4s r, mat4s m, vec4s c) {
   return glm_mat4_rmc(r.raw, m.raw, c.raw);
 }
 
+/*!
+ * @brief Create mat4 matrix from pointer
+ *
+ * @param[in]  src  pointer to an array of floats
+ * @param[out] dest matrix
+ */
+CGLM_INLINE
+mat4s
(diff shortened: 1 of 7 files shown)


recp:
@EasyIP2023 thanks for the PR,

What about `glm_mat2_make`, `glm_mat3_make`? also we have struct api it would be nice to add this to there too

include/cglm/struct/mat4.h
```diff
@@ -459,4 +459,17 @@ glms_mat4_(rmc)(vec4s r, mat4s m, vec4s c) {
...
 
+/*!
+ * @brief Given a 16 element float array create mat4 matrix
+ *
+ * @param[in]  src  pointer to float array
+ * @param[out] dest matrix
+ */
+CGLM_INLINE
+mat4s
+glm_mat4_(make)(float * __restrict src, mat4s dest) {
```
recp, line 470: You should use `glms_` namespace/prefix here not `glm_` which indicates inline array api.

recp (changes requested):
also docs/ and tests would be nice for new funcs if possible.

include/cglm/struct/mat4.h
EasyIP2023: Thanks changes made

recp:

PR #300, opened 2023-05-15
README: update meson option for running test

EasyIP2023:
Running on an already built build directory

```
meson configure -Denable_tests=true build/
```

Leads to

`ERROR: Unknown options: "enable_tests"`

Seems the meson option was updated


recp:

PR #302, opened 2023-05-15
mat3: add new function glm_mat3_make

EasyIP2023:
Function takes in a 9 element float array
and converts it into a mat3 matrix.


recp:

PR #303, opened 2023-05-15
io: fix test runner.c implicit declaration glm_arch_print

EasyIP2023:
Problem:
```
../test/runner.c:29:3: error: implicit declaration of function ‘glm_arch_print’ [-Werror=implicit-function-declaration]
   29 |   glm_arch_print(stderr);
      |   ^~~~~~~~~~~~~~
      |   glm_vec2_print
```

`glm_arch_print` is not available unless you add the `DEBUG` macro.


recp:
Thanks for the catch,

Let's don't define `DEBUG` manually, if should came from user / build conf. 

Another approach would be:

add the news funcs end of `io.h`:

```C
#define glm_arch_print(s) (void)s;
#define glm_arch_print_name(s) (void)s;
```

to make them noop like others. This should work.

EasyIP2023:
@recp fixed thanks

recp:
@EasyIP2023 thanks, the PR is merged 🚀

PR #321, opened 2023-06-29
quat: add new function glm_quat_make

EasyIP2023:
Not really necessary. Could just define the versor your self, but to be compatible with

https://glm.g-truc.net/0.9.9/api/a00722.html

and other portions of the API. Add new function that takes in a 4 element float array
and converts it into a quaternion.


recp:
@EasyIP2023 the PR is merged, thanks for your contributions 🚀

PR #323, opened 2023-07-02
test_quat: add more robust quat_make test

EasyIP2023:
Makes it so that it's easier to identify
the potential use case of function. Commit also
includes a fix to the struct/quat.h glms_quat_make comment.
Should be returning versors it's not
a void function.


recp:

PR #325, opened 2023-07-02
vec3: add new function glm_vec3_make

EasyIP2023:
Function takes in a float array. Array must be
at least of size 3 and converts it into
a 3D vector.


recp:

PR #326, opened 2023-07-02
vec4: add new function glm_vec4_make

EasyIP2023:
Function takes in a float array. Array must be
at least of size 4 and converts it into
a 4D vector.


recp:

PR #332, opened 2023-07-15
add missing mat2x# test and mat2x3 window headers

EasyIP2023:
Sorry @recp after looking through there were a few things missing


recp:
@EasyIP2023 thanks, the PR is merged :)

PR #341, opened 2023-07-31
more robust __builtin_assume_aligned detection

Akaricchi:
`__builtin_assume_aligned` is available since GCC 4.7, but `__has_builtin` was added much later. Check for the GCC version if `__has_builtin` is not available.

Users can also define `CGLM_HAVE_BUILTIN_ASSUME_ALIGNED` to either 1 or 0 to explicitly enable/disable the use of `__builtin_assume_aligned`. Meson will do it automatically (by performing a configure-time test).

### DIFF
--- a/include/cglm/types.h
+++ b/include/cglm/types.h
@@ -32,13 +32,27 @@
 #  define CGLM_ALIGN_MAT CGLM_ALIGN(16)
 #endif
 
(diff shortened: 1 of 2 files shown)


recp:
@Akaricchi thanks for your contributions, the PR is merged 🚀

PR #345, opened 2023-08-29
Fixed glms_mat4_ins3 bug.

telephone001:
Pat4_ins3 should copy mat3s to upper left mat4s. Previously, it just created an uninitialized mat4s r and put the mat3 into that and returned it. Now it takes in a mat4s argument and copies the mat3s to that.


recp:
Hi @telephone001,

Nice catch, many thanks, the PR is merged 🚀

PR #346, opened 2023-09-04
eqv, eq, fill for ivec2 and ivec3, documentation fixes

duarm:
- New `glm_ivec4_print`, `glm_ivec2_print` functions
- New `eqv`, `eq`, `fill` functions for `ivec3` and `ivec2`
- Adding missing `glm_vec2_print` to the docs.
- Adding missing call func `glmc_vec2_fill`, `glmc_vec2_eq`, `glmc_vec2_eqv`
- Changed wording on the print documention to a generic "vector" and "matrix" instead of the type, to be more inline with how it's worded everywhere else.


recp:
@duarm the PR is merged, thanks for your contributions 🚀 It would be nice to have ivec4 versions of the new functions :)

gottfriedleibniz:
Would it be possible to update `glm_ivec2_fill` and `glm_ivec3_fill` to explicitly cast those values? For example, compiling with MSVC `/W2` (or above) will throw:

```
cglm\include\cglm\ivec2.h(219,17): warning C4244: '=': conversion from 'float' to 'int', possible loss of data 
cglm\include\cglm\ivec3.h(227,24): warning C4244: '=': conversion from 'float' to 'int', possible loss of data 
```

recp:
@gottfriedleibniz wow my mistake, those parameters must be int as base type of vector, it is fixed at: https://github.com/recp/cglm/commit/126f809dae76d51f1527cf19627d7c08b02ed108

gottfriedleibniz:
If that's the case, `glmc_ivec{2,3}_fill` also require updating.

recp:
done :)

duarm:
missed that too, sorry

PR #347, opened 2023-09-06
No relative include

duarm:
This one is rather big but simple. Remove all relative includes, in the `#include "header.h"` or `#include "../header.h"` style, in favor of `#include <cglm/header.h>`

This makes the project more flexible to files moving, a wrong header inclusion becomes very unlikely, provides more information about what is being included and is more consistent (consumers include each header the exact same way as the library)

Only tested with cmake, skipped the test folder because it's not really necessary


recp:
Hi @duarm,

I think we must stick to relative headers since it guarantees that all headers come from the same release and source, ensuring consistent integration and reducing the risk of unintended cross-references.

When user set `header search path` for cglm, all public headers can be accessed as <cglm/***.h> as your proposal. 

Sorry. 

More feedbacks would be appreciated

duarm:
> Sorry.
> 
> More feedbacks would be appreciated

No problem, it's just a suggestion and it took 2 minutes.

> I think we must stick to relative headers since it guarantees that all headers come from the same release and source, ensuring consistent integration and reducing the risk of unintended cross-references.

I don't think this "ensures" anymore than a normal <> include, the compiler will always look for the -I flags first, then the system's path.

recp:
> I don't think this ensures anymore than a normal <> include, the compiler will always look for the -I flags first, then the system's path.

this is may be true for <> also would make things easier in src folder in large projects,  but since all headers are in same place; using relative path in public headers seems more appropriate for me. Since there is no need to build cglm ( can be used inline only ), someone can include **cglm** into a project with drag-drop of sources. In this case there is no need to set -I, it can be used directly, otherwise <> would make things a bit harder for scenarios like this. Consider that if cglm is installed to global place like /usr/local then copy-pasted sources may use /usr/local *( because of <cglm/...> if -I is set to user local )* which may be a different version...

PR #350, opened 2023-10-11
adding glm_vec2_center

duarm:
new glm_vec2_center function, there's already a vec3 version, this will be useful for when we implement 2d aabb


recp:
Hi @duarm,

Thanks, is it possible to add this function to struct api (include/cglm/struct/) and docs/ too, otherwise I'll do it later after merge the PR.

duarm:
I can do it no prob

duarm:
done, I've already added the docs to docs/source/vec2.rst, couldn't find a docs for the struct api, if there is such a thing

recp:
@duarm many thanks, the PR is merged 🚀

> couldn't the struct api docs, if there is such a thing

Since it is same job, there is no need to duplicate docs

PR #360, opened 2023-12-01
meson: don't build by default if used in a subproject

v1993:
This allows to do the following in a project using cglm as a subproject and avoid building cglm's standalone library if you only use inline versions of functions:

```meson
cglm_dep = dependency(
  'cglm',
  default_options: [ 'install=false' ]
).partial_dependency(
  compile_args: true,
  includes: true,
)

# Needed since it's won't be pulled in automatically now
libm_dep = meson.get_compiler('c').find_library('m', required: false)
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -95,6 +95,7 @@ cglm_dep = declare_dependency(
     dependencies : cglm_deps,
     compile_args : cglm_args,
     include_directories : cglm_inc,
+    build_by_default: not meson.is_subproject(),
     version : meson.project_version()
 )
 


recp:
@v1993 thanks for your contributions, 

Should we add this to meson section of README, Docs to let users know how to use inline-only version of **cglm** ?

v1993:
I think a note in readme would be nice, yes. This works with system-wide cglm as well.

kzhsw:
[CI ](https://github.com/recp/cglm/actions/runs/7070293843/job/19246847085) seems broken, any idea?
> meson.build:93:11: ERROR: declare_dependency got unknown keyword arguments "build_by_default"

v1993:
> [CI ](https://github.com/recp/cglm/actions/runs/7070293843/job/19246847085) seems broken, any idea?
> 
> > meson.build:93:11: ERROR: declare_dependency got unknown keyword arguments "build_by_default"

That's what I get for making this PR from web interface and not double-checking :facepalm:. The new line was supposed to go in `library` call just above, not `declare_dependency`.

recp:
https://github.com/recp/cglm/commit/abb71a8e32ff56314ce70edc0bb92b332efb704a should fix the issue, @v1993 right?

v1993:
> [abb71a8](https://github.com/recp/cglm/commit/abb71a8e32ff56314ce70edc0bb92b332efb704a) should fix the issue, @v1993 right?

Yep, that seems to be correct. Thanks for getting this sorted out!

recp:
@v1993 thanks 👍

PR #361, opened 2023-12-02
wasm: prefer pmin/pmax

myfreeer:
According to [emscripten](https://emscripten.org/docs/porting/simd.html) and [v8](https://github.com/v8/v8/blob/b6520eda5eafc3b007a5641b37136dfc9d92f63d/src/compiler/backend/x64/code-generator-x64.cc#L2661-L2699), `[f32x4|f64x2].[min|max]` compiles to much more instructions than `[f32x4|f64x2].[pmin|pmax]`. It is defined in [spec](https://github.com/WebAssembly/spec/blob/main/proposals/simd/SIMD.md#floating-point-min-and-max) that the difference between pmin/pmax and min/max is NaN-propagating behavior, and the equivalent to the x86 `_mm_min_ps`/`_mm_max_ps` is pmin/pmax in [v8](https://github.com/v8/v8/blob/b6520eda5eafc3b007a5641b37136dfc9d92f63d/src/compiler/backend/x64/code-generator-x64.cc#L2740-L2747). This should make functions with min/max faster on webassembly, and align with the existing behavior with x86 sse.

### DIFF
--- a/include/cglm/vec4.h
+++ b/include/cglm/vec4.h
@@ -649,7 +649,7 @@ glm_vec4_maxadd(vec4 a, vec4 b, vec4 dest) {
 #if defined(__wasm__) && defined(__wasm_simd128__)
   glmm_store(dest, wasm_f32x4_add(
           glmm_load(dest),
-          wasm_f32x4_max(glmm_load(a), glmm_load(b))));
+          wasm_f32x4_pmax(glmm_load(a), glmm_load(b))));
 #elif defined( __SSE__ ) || defined( __SSE2__ )
   glmm_store(dest, _mm_add_ps(glmm_load(dest),
                               _mm_max_ps(glmm_load(a),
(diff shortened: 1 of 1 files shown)


recp:
@myfreeer many thanks for contributions for WASM support and its stability, the PR is merged  🚀

gottfriedleibniz:
Should we aim for consistent Nan-propagating behaviour here? The Intel ISA states:

> If only one value is a NaN (SNaN or QNaN) for this instruction, the second operand (source operand), either a NaN or a valid floating-point value, is written to the result.

From the V8 link above:

> The maxps instruction doesn't propagate NaNs and +0's in its first operand

Therefore to match `_mm_max_ps(a, b)`: `wasm_f32x4_pmax(a, b)` should be `wasm_f32x4_pmax(b, a)`. For example, consider:

```c
vec4 a, b, c;

glm_vec4_broadcast(NAN, a);
glm_vec4_broadcast(0.0, b);
glm_vec4_maxv(a, b, c);

printf("%f %f %f %f\n", c[0], c[1], c[2], c[3]);
```

Similar logic applies to `min`.

recp:
@gottfriedleibniz many thanks for catching edge-cases like this and help to fix them. 

I think we can wrap min/max by `glmm_` e.g. `glmm_min()` and `glmm_max()` to make behaviors identical

Issue #322, opened 2023-07-02
Missing Handed / Clipspace bound implementations 


recp:
Hi,

I have noticed that there are several functions that seem to be implemented in a right-handed (**RH**) or NegatifZero-One (**ZO**) context only. 

Some of them:

- Quat to martix
- Matrix to Quat?
- Some other quat functions maybe
- `glm_vec3_rotate`
- 2D and 3D Affine functions 
- Frustum helpers 
- others maybe...

As cglm supports LH, RH, ZO and NO, it would be nice to identify and implement, improve existing  implementations across all contexts where applicable

Any help in this regard would be highly appreciated.

best,
r.

Issue #306, opened 2023-05-30
How do I generate a quaternion from euler angles?


badasahog:
I want to get a quaternion based on roll, pitch, and yaw, but as far as I can tell there is no function for this.

The only functions I could find were based on a unit vector and an angle (glm_quatv, glm_quat)

glm provides this functionality in the form of the glm::quat(glm::vec3) constructor

recp:
Hi @badasahog,

Sorry for the delay,

It could be nice to have this feature in **cglm** too. I had tried to update Euler with https://github.com/recp/cglm/pull/128 but couldn't finished that PR yet,

We can have this before that,

Any PR would be awesome otherwise I added to TODOs. 🙂

Nonesence999:
Yes It would be amazing I also need a quat structure for my project.

telephone001:
Sorry for answering very late, but for anyone else who needs help in the future, you could try making a function that rotates the quaternion by each axis. I guess you could also try and push it to the repository even though it might not be optimal

// example for rotating the quaternion by the y axis by an angle:

// y axis
vec3s axis = {0, 1, 0};

// creates a quaternion with the angle and axis
versors rotation = glms_quatv(angle, axis); 

// when multiplying this rotation with an angular position quaternion, it will apply the rotation to the angular position
angular_position = glms_quat_mul(rotation, angular_position);

recp:
@telephone001 many thanks, hope we add this asap 🚀

telephone001:
do you think glm_quate_xyz or glm_quat_euler_xyz would be a better name for this new function?

recp:
To convert Euler to Quat, I would use `glm_euler_[x|y|z]_quat` which is similar pattern used in **cglm** e.g. glm_mat4_quat() convert mat4 to quat. 

```
glm_SRC_DST()
```

telephone001:
Alright thanks for the response. I'll try and get this done soon

recp:
@telephone001 thanks 🤗

recp:
Thanks for they implementation a PR would be nice to see,

> it fails using meson webassembly for some reason.

https://github.com/recp/cglm/actions seems fine, may I ask what are the errors? Also can you pull latest commits and try again? there was an issue about https://github.com/recp/cglm/pull/360#issuecomment-1838122934 but fixed later.

telephone001:
I created a pr and most tests failed. The error is just that all tests fail on certain builds.

gottfriedleibniz:
[...]

PR #369, opened 2023-12-08
euler to quat functions

telephone001:
added all 6 combinations of glm_euler_xyz_quat. The tests work on my computer but they don't seem to work in the meson webassembly and the checks fail. I might be misreading or missing something though.


telephone001:
I'm thinking of:
1. putting this in euler.h since it seems like it will fit better there
2. using a parameter to control what order the euler rotations will take place in the quaternion
3. If the xyz combinations have to happen in the name, using a macro function to create all tests for all combinations of xyz
4. I'll also add glm_euler_by_order + glm_mat4_quat in tests as another test.

I'd like some feedback of these changes

include/cglm/euler.h
```diff
@@ -448,4 +454,168 @@ glm_euler_by_order(vec3 angles, glm_euler_seq ord, mat4 dest) {
...
+void
+glm_euler_xyz_quat(versor q, vec3 angles) {
+  float xs = sinf(angles[0] / 2.0f);
+  float xc = cosf(angles[0] / 2.0f);
+
+  float ys = sinf(angles[1] / 2.0f);
+  float yc = cosf(angles[1] / 2.0f);
+  
+  float zs = sinf(angles[2] / 2.0f);
+  float zc = cosf(angles[2] / 2.0f);
```
recp, line 475: I know new compiler and C versions can handle this but I prefer to do declarations beginning of a scope block:

```C
  float xs, xc, ys, yc, zs, zc;

  xs = sinf(angles[0] * 0.5f);
```

also even compilers do optimizations it would be nice to multiply with 0.5 rather than dividing by 2.

recp:
@telephone001 many thanks for the work,

just a reminder, as mentioned: https://github.com/recp/cglm/issues/306#issuecomment-1848052488 better to double check this before merge

docs/euler.rst: 

```
Rotation Conveniention
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Current *cglm*'s euler functions uses these convention:

* Tait–Bryan angles (x-y-z convention)
* Intrinsic rotations (pitch, yaw and roll).
  This is reserve order of extrinsic (elevation, heading and bank) rotation
* Right hand rule (actually all rotations in *cglm* use **RH**)
* All angles used in *cglm* are **RADIANS** not degrees
```

In the future euler api can be improved even may provide global/extrinsic too but currently as docs states we must stick above constraints.

telephone001:
Currently my functions work like doing glm_quat_mul 3 times  with the axis vectors. But when compared to the glm_euler_by_order, all the values are negative compared to using glm_euler_by_order.

I pushed just to give an update. This commit is messy because I'm in the process of debugging it and debugging tests in euler.c is harder than in quat.h. 

Currently i'm not really getting what extrinsic and intrinsic angles are. I based my calculations thinking that xyz corresponds to yaw pitch roll.

recp:
> Currently i'm not really getting what extrinsic and intrinsic angles are.

https://dominicplein.medium.com/extrinsic-intrinsic-rotation-do-i-multiply-from-right-or-left-357c38c1abfd explain and shows extrinsic and intrinsic graphically (animated) which could help to understand better. 

Tait-Bryan vs. Classic,
Intrinsic vs. Extrinsic rotations...:

https://en.wikipedia.org/wiki/Euler_angles
https://danceswithcode.net/engineeringnotes/rotations_in_3d/rotations_in_3d_part1.html 

--- 

I'll check glm_euler_by_order to validate glm_euler_by_order (..., GLM_EULER_XYZ, ..) is same as glm_euler_xyz()... asap

telephone001:
The problem might be in my tests as there are some xyzs that are in the wrong spots. I'll try and get that fixed up

telephone001:
Currently all my tests pass when only checking multiplication of quaternion axises. I'm pretty sure glm_euler_by_order is the same as glm_euler_xyz and its variants because it returns the same results when testing. I'm thinking the error might be in the mat4 ->quat or euler->mat4.

Also, I remember not being able to make different tester functions even when I declared and entered them. I did the exact same thing I did when making the tests in testquat.h. they worked in testquat but not euler.c. Could I possibly be missing something?

recp:
@telephone001 many thanks, I'll try to take a deeper look asap

telephone001:
when test_assert_quat_eq_abs used, all tests pass. But I still want the directions of my function to be right and consistent with the other glm functions

gottfriedleibniz:
That may be a bit difficult given the nature of `glm_mat4_quat`. 

Another question is whether the ordering of parameters be flipped? Out pointers (i.e., "dest") parameters are usually the last parameters to a function, e.g., `glm_euler_by_order`.

telephone001:
Oh yeah I'll make the destination the last parameter if thats needed. Some of the quat functions have the output as the first parameter though when they are creating a new quaternion.

telephone001:
I'm thinking of either removing the quat tests and merging then dealing with mat4 in another request or making another pull request and fixing mat4 there then after its done, go back here and merge.

telephone001:
I also figured out how to make the tests separate. I'm gonna be busy for a while but later I'm gonna try and make better tests for euler_xyz and possibly fix them

telephone001:
Oh I forgot that could happen. Euler_xyz might be fine. I tried applying the mat4 and quaternion to the same vec3 and they both returned the same result. I'm pretty sure mine computes the longer path if angle > 180* because it always gets the same results as rotating by 3 axis quaternions with the angles just plugged in. I'll try and fix it soon

recp:
@telephone001 thanks 👍

I cant validate that mat4_quat always gives short path, I'll check it when I available asap.

telephone001:
@recp 
wait, I just realized that I was comparing the quaternions with eachother and a negative quaternion expresses the same rotations as a positive quaternion just that the axis and angles have been reversed. All I would need to do is use the abs test like @gottfriedleibniz said.

telephone001:
@recp 
I'm thinking that instead of forcing the quaternion to always be the shortest path always, we can just change the quaternion to the short path when needed in interpolation. I think the path it takes only matters in interpolation because if you just want to apply a quaternion to a position, you just use the hamiltonian product and the position gets oriented instantly so I'm gonna take gottfriedleibniz's advice and just use the abs testers

recp:
@telephone001, are we ready to merge this PR? I'll check the implementation this weekend.

One question:

since **cglm** supports LH operations too, I guess we may need to provide LH versions too e.g. `glm_euler_xyz_quat_lh()`.

One option is to implement this as RH then LH asap to prevent this become giant PR

see: https://github.com/recp/cglm/issues/322

telephone001:
@recp
Oh yeah I thought you were reviewing my pr last week. Sorry I couldn't respond sooner.
I'll change all these to RH and then push again. Then later do it in left hand

recp:
> Oh yeah I thought you were reviewing my pr last week

Sorry for the delay :( 

> Do I need to rename them to glm_euler_xyz_quat_rh()?

That would be awesome. I think we can move LH/RH related implementations into `handed` folder similar to  `clipspace` and wrap them by options e.g. `#ifdef CGLM_FORCE_LEFT_HANDED`. See `cam.h` which can help to implement **LH**, **RH** skeleton ( for instance files in separate *handed* folder and wrapper in euler.h ). 

If you dont enough time I can merge this PR and add _rh() suffixes and move them to `handed` folder ( if there is a better name we can continue with it )

> I'll get it in left hand in another PR.

👍

telephone001:
Should I make euler_to_quat_rh.h and euler_to_quat_lh.h in that folder?
Also instead of clipspace and handed folders, I just have 1 handed folder and put all handed stuff in that folder?

telephone001:
currently I have a handed folder and also the rh stuff in that folder. I didn't know what to name the macro that controls righthandedness so I just used CGLM_FORCE_LEFT_HANDED

recp:
> Should I make euler_to_quat_rh.h and euler_to_quat_lh.h in that folder?

euler_rh.h/euler_lh.h would be enough but euler_to_quat_rh.h is also ok. 

> Also instead of clipspace and handed folders, I just have 1 handed folder and put all handed stuff in that folder?

`clipspace` keeps stuff related clipspace and/or lh/rh... I think `handed` can just keep stuff only lh/rh.

using `CGLM_FORCE_LEFT_HANDED` is right path to go for now 👍

recp:
@telephone001 the PR is merged, many thanks for your contributions 🚀 

Let's continue with separate PR for LH versions and keep track bugs and improvements for a while :)

PR #377, opened 2023-12-28
euler_to_quat_lh also some euler_to_quat bugfixes

telephone001:
Creates euler_to_quat_lh
makes euler_to_quat_rh and lh struct functions
fixes up euler_to_quat documentation

made it early so I remember about it

### DIFF
--- a/include/cglm/euler.h
+++ b/include/cglm/euler.h
@@ -460,8 +460,8 @@ glm_euler_by_order(vec3 angles, glm_euler_seq ord, mat4 dest) {
  * @brief creates NEW quaternion using rotation angles and does
  *        rotations in x y z order (roll pitch yaw)
  * 
- * @param[out]  q     quaternion
- * @param[in]   angle angles x y z (radians)
+ * @param[in]   angles angles x y z (radians)
+ * @param[out]  dest   quaternion
  */
 CGLM_INLINE
 void
@@ -477,8 +477,8 @@ glm_euler_xyz_quat(vec3 angles, versor dest) {
  * @brief creates NEW quaternion using rotation angles and does
  *        rotations in x z y order (roll yaw pitch)
  * 
- * @param[out]  q     quaternion
- * @param[in]   angle angles x y z (radians)
+ * @param[in]   angles angles x y z (radians)
+ * @param[out]  dest   quaternion
  */
 CGLM_INLINE
 void
@@ -494,8 +494,8 @@ glm_euler_xzy_quat(vec3 angles, versor dest) {
  * @brief creates NEW quaternion using rotation angles and does
  *        rotations in y x z order (pitch roll yaw)
  * 
- * @param[out]  q     quaternion
- * @param[in]   angle angles x y z (radians)
+ * @param[in]   angles angles x y z (radians)
+ * @param[out]  dest   quaternion
  */
 CGLM_INLINE
 void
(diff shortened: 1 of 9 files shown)


telephone001:
For left handed operations in the future, do you think it will be nessesary to have CGLM_EULER_CONTROL_RH_BIT. It fits with the style but for euler, there is only left or right so we can just use CGLM_FORCE_LEFT_HANDED in the code.

#define CGLM_EULER_CONTROL_RH_BIT (1 << 0) /* LEFT_HANDED, For DirectX, Metal, Vulkan */
#define CGLM_EULER_CONTROL_LH_BIT (1 << 1) /* RIGHT_HANDED, For OpenGL, default in GLM */

#ifdef CGLM_FORCE_LEFT_HANDED
#  define CGLM_CONFIG_EULER_CONTROL
#else
#  define CGLM_CONFIG_EULER_CONTROL
#endif

telephone001:
Also, the only difference between euler to quat rh vs lh is that the zsin part is negative. This might be helpful info for someone I think. 
I was thinking of doing 
#ifdef CGLM_FORCE_LEFT_HANDED
#define CGLM_EULER_ZS -sinf(angles[2] * 0.5f)
#else
#define CGLM_EULER_ZS sinf(angles[2] * 0.5f)
and using that to choose between lh and rh. Currently idk how to do it so there is an option for both to be there.

recp:
Many thanks @telephone001 for the work!

For now, I think it is enough to continue with `CGLM_FORCE_LEFT_HANDED` for LH/RH separation. 


> Also, the only difference between euler to quat rh vs lh is that the zsin part is negative. This might be helpful info for someone I think.
> I was thinking of doing
> #ifdef CGLM_FORCE_LEFT_HANDED
> #define CGLM_EULER_ZS -sinf(angles[2] * 0.5f)
> #else
> #define CGLM_EULER_ZS sinf(angles[2] * 0.5f)
> and using that to choose between lh and rh. Currently idk how to do it so there is an option for both using that.

that would makes it easy to maintain. But we can rewrite xyz:

```C
  // rH
  dest[0] = xc * ys * zs + xs * yc * zc;
  dest[1] = xc * ys * zc - xs * yc * zs;
  dest[2] = xc * yc * zs + xs * ys * zc;
  dest[3] = xc * yc * zc - xs * ys * zs;

  // Lh
  dest[0] = xs * yc * zc - xc * ys * zs;
  dest[1] = xc * ys * zc + xs * yc * zs;
  dest[2] = xs * ys * zc - xc * yc * zs;
  dest[3] = xc * yc * zc + xs * ys * zs;
```

by keeping this part same:

```C
  xs = sinf(angles[0] * 0.5f); xc = cosf(angles[0] * 0.5f);
  ys = sinf(angles[1] * 0.5f); yc = cosf(angles[1] * 0.5f);
  zs = sinf(angles[2] * 0.5f); zc = cosf(angles[2] * 0.5f);
```

not sure yet but maybe we can apply some simd on these in the future and separate implementations would make things easier. Or maybe negating single angle then applying same implementation will be better choice. 

For now we can continue as your implementation. Adding `Also, the only difference between euler to quat rh vs lh is that the zsin part is negative. This might be helpful info for someone I think.` to somewhere (e.g. top of the header) would be helpful for the future.  

I guess we are ready to merge the PR?

telephone001:
yeah I think we can merge now

recp:
@telephone001 thanks for your contributions, the PR is merged 🚀

PR #379, opened 2024-01-09
simd: min / max helpers

recp:
New SIMD helpers:

- `glmm_min(a, b)`
- `glmm_max(a, b)`

simd helpers to match x86, neon and wasm behaviors ( without compromising performance ). I reversed `a` and `b` for wasm as @gottfriedleibniz suggested: https://github.com/recp/cglm/pull/361

```C
static inline __m128 glmm_min(__m128 a, __m128 b) { return _mm_min_ps(a, b); }
static inline __m128 glmm_max(__m128 a, __m128 b) { return _mm_max_ps(a, b); }

static inline float32x4_t glmm_min(float32x4_t a, float32x4_t b) { return vminq_f32(a, b); }
static inline float32x4_t glmm_max(float32x4_t a, float32x4_t b) { return vmaxq_f32(a, b); }

static inline glmm_128 glmm_min(glmm_128 a, glmm_128 b) { return wasm_f32x4_pmin(b, a); }
static inline glmm_128 glmm_max(glmm_128 a, glmm_128 b) { return wasm_f32x4_pmax(b, a); }
```

@gottfriedleibniz, @myfreeer a review would be awesome. 

In NEON there are also `vminnmq_f32()`/`vmaxnmq_f32()` which are similar to `vminq_f32()`/`vmaxq_f32()`, not sure which is best to use to mimic `_mm_min_ps/_mm_max_ps()`.

I think since we dont expect NaNs, 'maximum performance' is preferred, any ideas?

### DIFF
--- a/include/cglm/simd/arm.h
+++ b/include/cglm/simd/arm.h
@@ -56,11 +56,9 @@ glmm_float32x4_init(float x, float y, float z, float w) {
 #define glmm_float32x4_SIGNMASK_NPNP glmm_float32x4_init(-0.f,  0.f, -0.f,  0.f)
 #define glmm_float32x4_SIGNMASK_NPPN glmm_float32x4_init(-0.f,  0.f,  0.f, -0.f)
 
-static inline
-float32x4_t
-glmm_abs(float32x4_t v) {
-  return vabsq_f32(v);
-}
+static inline float32x4_t glmm_abs(float32x4_t v)                { return vabsq_f32(v);    }
+static inline float32x4_t glmm_min(float32x4_t a, float32x4_t b) { return vminq_f32(a, b); }
+static inline float32x4_t glmm_max(float32x4_t a, float32x4_t b) { return vmaxq_f32(a, b); }
 
 static inline
 float32x4_t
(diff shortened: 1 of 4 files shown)


kzhsw:
For wasm, is it worth to give user an option to use the instructions from [relaxed-simd](https://github.com/WebAssembly/relaxed-simd) proposal for maximum performance? It is currently in [Phase 4](https://github.com/WebAssembly/proposals), but only supported in limited environment (currently chrome 114 and wasmtime 15).

myfreeer:
Agreed with maximum performance, maybe some continous benchmark would help to mesure performance impact of pr or commit.

gottfriedleibniz:
With this PR, functions like `glm_vec4_maxv` can also be simplified.

> In NEON there are also vminnmq_f32()/vmaxnmq_f32()

Only for ARMv8+ and it follows IEEE behaviour for NaN handling.

> match x86, neon and wasm behaviors

Trying to match the behaviour here is a bit annoying. For NEON to emulate `_mm_max_ps` and `glm_vec4_maxv` without intrinsics, it would require something along the lines of `vbslq_f32(vcltq_f32(b, a), a, b)` (verification required)

recp:
@kzhsw thanks for sharing that, in the future to provide max performance some options can be provided 👍 I'm not wasm expert so cant talk much for now :) 

> maybe some continous benchmark would help to mesure performance impact of pr or commit.

@myfreeer I liked the benchmark idea, TODOs 🤗

> With this PR, functions like glm_vec4_maxv can also be simplified.

@gottfriedleibniz thanks for your feedbacks the commit must do that: https://github.com/recp/cglm/pull/379/commits/6d8dd42ac26e361ccbfaf494eb5cb17a035fb52f 

> Trying to match the behaviour here is a bit annoying. For NEON to emulate _mm_max_ps, and glm_vec4_maxv without intrinsics, it would require something along the lines of vbslq_f32(vcgtq_f32(a, b), a, b) (verification required) ... which is a bit overkill.

actually just tried to match only simd behaviors where possible without bringing extra overhead but lets ignore it. 👍

Going to merge this, hope we will make it better over time.

Issue #382, opened 2024-02-03
Struct API for ivec2s, ivec3s, and ivec4s


tarhses:
Hello,

I just started to use this lib and noticed that there aren't any struct API function counterparts of `glm_ivec2_*`, `glm_ivec3_*`, etc. Is this intentional?

I started implementing cglm/struct/ivec2.h on my side and would gladly contribute if welcome :smiley:

Anyway thanks for the hard work!

recp:
Hi @tarhses,

This is discussed a bit here: https://github.com/recp/cglm/discussions/370#discussioncomment-7806749

> I started implementing cglm/struct/ivec2.h on my side and would gladly contribute if welcome 😃

That would be awesome! 

> Anyway thanks for the hard work!

Many thanks 😊

recp:
@tarhses the PR is merged, many thanks for your contributions, we can close the issue now

PR #383, opened 2024-02-07
Implement struct API for ivec2, ivec3, and ivec4

tarhses:
Related to #382.

This PR's goal is to implement struct API for ivec2, ivec3, and ivec4. I tried to closely follow what was already done with vec2/3/4 struct API and I also added missing print functions.


recp:
@tarhses thanks for your contributions, the PR is merged 🚀

PR #386, opened 2024-02-10
Add additional CI via GitHub Actions.

waywardmonkeys:
This adds CI via GitHub Actions for various targets, some of which haven't been tested via CI before.

* Autotools (Mac and Linux).
* cmake on iOS using Xcode.
* cmake on macOS using ninja.
* cmake on Linux using ninja, with both clang and gcc checked.
* cmake on Windows using MSVC (x64 and x86, Clang, ARM64, Windows Store).
* Documentation via Sphinx.
* Meson on macOS and Linux.
* MSBuild on Windows (x64 and x86).
* Swift on macOS and Linux.

There are some things that could be improved in subsequent PRs:

* Remove Appveyor config and README link.
* Remove TravisCI config. I left this as it contains the old config for coverage.
* Add code coverage again.
* Fix building ARM and ARM64 targets via MSBuild.
* Fix building ARM via cmake on Windows.
* Have Sphinx fail on warnings (but keep going), once Sphinx warnings are fixed.
* Add tests for Swift and run them.
* Add emscripten builds + tests, absorbing the other 2 workflow configurations.
* Make sure SIMD is built and tested where available.
* Add msys2 builds.
* Add Linux ARM64 builds.


waywardmonkeys:
You can see all of the CI checks here in this PR as they're prefixed with "CI /"

recp:
@waywardmonkeys many thanks for CI support 🤗 the PR is merged 🚀

PR #387, opened 2024-02-10
glm_ease_back_inout: Use float constant for float var.

waywardmonkeys:


recp:
@waywardmonkeys thanks, the PR is merged 🚀

PR #391, opened 2024-02-12
ci: Add ARM64 UWP for WindowsStore.

waywardmonkeys:
This is one of the core vcpkg targets and I'm trying to enable the cglm port there to support UWP, so we should have it in the CI here.


waywardmonkeys:
The PR for vcpkg: https://github.com/microsoft/vcpkg/pull/36704

PR #392, opened 2024-02-19
aabb2d_zero, aabb2d_diagonal and aabb2d_size

duarm:
-  adds `_aabb2d_zero`.

### Problem

Currently `aabb2d_size` returns `vec2_distance(min, max)`, which would be the diagonal of the aabb, if you want to use the size to resolve collisions, you need to work backwards from what aabb2d_size does when it calls vec2_distance. 

### Solution

- renames `_size` -> `_diag`. And provides a macro for compatibility
- adds a `_sizev` function which outputs a vec2, which is basically glm_vec2_sub(max, min, out).

image

Looking for feedback if this is desirable, maybe come up with another name to avoid a breaking change on _size?

### DIFF
--- a/docs/source/aabb2d.rst
+++ b/docs/source/aabb2d.rst
@@ -17,6 +17,10 @@ convert it before and after calling a cglm aabb2d function.
 Table of contents (click to go):
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+Macros:
+
+1. :c:func:`glm_aabb2d_size`
+
 Functions:
 
 1. :c:func:`glm_aabb2d_copy`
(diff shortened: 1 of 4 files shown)


duarm:
I should also split _zero into another PR

recp:
Hi @duarm,

Sorry for the delay, I like the your new interpretation of `_size` and `_diagonal`/`_diag` we can go in this way. This will break existing codebase so keeping and deprecating `_size` for backward compatibility, introduce `_sizev` ( v/vector suffix ) for sizes as vector and `_diagonal`/`_diag` for size of diagonal would be an option. 

Feedbacks are welcome

duarm:
size -> diag and new _sizev function, made a macro for _size to _diag. Should we already do the same for aabb?

duarm:
size -> diag and new _sizev function, made a macro for _size to _diag. Should we already do the same for aabb, or should I open another PR?

recp:
> Should we do the same for aabb, or should I open another PR?

Yes other PR would be better, thanks 👍

recp:
@duarm the PR is merged, thanks for your contributions 🚀

Issue #385, opened 2024-02-09
Non-square matrix * column vector multiplication implemented as row vector * matrix multiplication 


Error-mdl:
The matrix-vector multiplication functions for non-square types (mat4x3, mat3x4, mat2x4, etc) incorrectly take the dot product of each column of the matrix with the vector instead of the row, which is vector * matrix not matrix * vector. This also means the functions take and output the wrong dimension of vector.  For example, the mat4x3 multiplication:

[https://github.com/recp/cglm/blob/45134b126548206b669154fb4c62ab8c0fc81206/include/cglm/mat4x3.h#L151](https://github.com/recp/cglm/blob/45134b126548206b669154fb4c62ab8c0fc81206/include/cglm/mat4x3.h#L151)

A mat4x3 is a matrix of 4 columns and three rows:
```
| m00 m10 m20 m30 | 
| m01 m11 m21 m31 |
| m02 m12 m22 m32 |
```
A valid matrix-vector multiply on this matrix would take a vec4 and output a vec3 where the first component of the vec3 is the dot product of the first row of the matrix with the vector:
```
d0 = (m00 * v0) + (m10 * v1) + (m20 * v2) + (m30 * v3)
``` 
instead, the `glm_mat4x3_mulv` takes a vec3 and outputs a vec4, where the first component of the output vec4 is
```
d0 = (m00 * v0) + (m01 * v1) + (m02 * v2)
```
[https://github.com/recp/cglm/blob/45134b126548206b669154fb4c62ab8c0fc81206/include/cglm/mat4x3.h#L154](https://github.com/recp/cglm/blob/45134b126548206b669154fb4c62ab8c0fc81206/include/cglm/mat4x3.h#L154)

The square matrix types correctly dot the rows of the matrix with the vector to get each component of the output:
[https://github.com/recp/cglm/blob/45134b126548206b669154fb4c62ab8c0fc81206/include/cglm/mat4.h#L407](https://github.com/recp/cglm/blob/45134b126548206b669154fb4c62ab8c0fc81206/include/cglm/mat4.h#L407)

EasyIP2023:
@Error-mdl

Seems like the main issue here is perspective of what the matrix looks like under the hood.

@recp We should probably document this. What do you think?

You say
https://github.com/recp/cglm/blob/45134b126548206b669154fb4c62ab8c0fc81206/include/cglm/mat4x3.h#L151
```
A valid matrix-vector multiply on this matrix would take a

vec4 and output a vec3

[...]

PR #403, opened 2024-03-30
mat2x3: fix multiplication functions

EasyIP2023:
Believe this is what we want. Per issue: https://github.com/recp/cglm/issues/385

TODO:

- [x] Update docs
- [x] Re-review comments
- [x] Re-review tests

### DIFF
--- a/docs/source/mat2x3.rst
+++ b/docs/source/mat2x3.rst
@@ -23,6 +23,16 @@ Functions:
 #. :c:func:`glm_mat2x3_transpose`
 #. :c:func:`glm_mat2x3_scale`
 
+Represented
+~~~~~~~~~~~
+
+.. csv-table:: mat2x3
+   :header: "", "column 1", "column 2"
+
+   "row 1", "m00", "m10"
+   "row 2", "m01", "m11"
+   "row 3", "m02", "m12"
+
 Functions documentation
 ~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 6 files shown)


recp:
Hi @EasyIP2023 ,

**Thanks for fixing it. I had forgotten them :(** 

You are accessing memory randomly which is not desired in **cglm**. Compiler may optimize it but compilers may not have same capabilities 

Desired access:

```C
  float a00 = m1[0][0], a01 = m1[0][1], a02 = m1[0][2],
        a10 = m1[1][0], a11 = m1[1][1], a12 = m1[1][2]
```

```
  COL0: a00 = m1[0][0], a01 = m1[0][1], a02 = m1[0][2],
  COL1: a10 = m1[1][0], a11 = m1[1][1], a12 = m1[1][2]
```

Cache locality is important even compiler may do this for us, given hint is desired style. Same for setting `dest[]`:

```C
dest[0][0] = ...
dest[0][1] = ...

dest[1][0] = ...
dest[1][1] = ...
```

Thanks

EasyIP2023:
@recp Now it makes perfect since why. Okay Changing now!

Do you think I should also do the same in `_mulv` and `_mulv`?

Or would assigning variables then do math take longer then accessing elements randomly and do math?

EasyIP2023:
@recp 

Just read other message will remove `_mulv`. Until more feedback given.

recp:
> Do you think I should also do the same in _mulv and _mulv?

I'm not sure if there is a benefit to move vector on stack but not matrix, access both on array or both on stack I think but not sure. Probably compiler will do the job but asm output can be compared. 

Coding style:

Declare first then init pls. ( except array init, const init ) 

```C
float v0 = v[0], v1 = v[1];
```

to 

```C
float v0, v1;

v0 = v[0];
v1 = v[1];
```

for similar places... 

> Just read other message will remove _mulv. Until more feedback given.

No, _mulv remain exist just waiting existence of _vmul.

EasyIP2023:
@recp

lol

Meant `_vmul` not `_mulv`.

EasyIP2023:
@recp 

This branch should be all cleaned up and ready for another review now.

recp:
@EasyIP2023 many thanks, the PR is merged 🚀

recp:
EDIT:

> float v0 = v[0], v1 = v[1];

This seems better fit existing code style, especially reading items from matrices, vectors... My actual intention was more general. Maybe variables that are not reading func params.

Issue #404, opened 2024-03-30
`I` macro is used as variable name in new vec_reflect / vec_refract functions


olifre:
Including `complex.h` from standard C (see https://en.cppreference.com/w/c/numeric/complex ) defines the macro `I`, which causes issues with the variable name `I` used starting with CGLM 0.9.3. 

See for example:
```
In file included from ../src/util/compat.h:13,
                 from ../src/taisei.h:13,
                 from ../src/boss.c:9:
../subprojects/cglm/include/cglm/vec2.h:725:23: error: expected ‘)’ before ‘__extension__’
  725 | glm_vec2_reflect(vec2 I, vec2 N, vec2 dest) {
      |                       ^
In file included from ../subprojects/cglm/include/cglm/cglm.h:12,
                 from ../src/util/glm.h:20,
                 from ../src/boss.c:17:
../subprojects/cglm/include/cglm/vec2.h:725:26: error: expected ‘;’, ‘,’ or ‘)’ before ‘vec2’
  725 | glm_vec2_reflect(vec2 I, vec2 N, vec2 dest) {
      |                          ^~~~
```
in one program using CGLM and including `complex.h` before.

recp:
Hi @olifre,

Wow, thanks for reporting this, I tried to fix this at https://github.com/recp/cglm/pull/406 and new Release will be soon.

recp:
@olifre the PR is merged I think the issue was dissolved. v0.9.4 will be released soon which will contain this bugfix.

recp:
The issue is fixed at v0.9.4: https://github.com/recp/cglm/releases

Thanks for reporting this

PR #406, opened 2024-03-31
dont use I macro defined in standard

recp:
adresses https://github.com/recp/cglm/issues/404, https://github.com/taisei-project/taisei/issues/381

I tried to fix this quickly hope there is no typo.


olifre:
Thanks for the swift action! 
I can confirm applying this patch to cglm, I can compile e.g. taisei fine again.

PR #409, opened 2024-03-31
docs: expand wy-nav-content width to edge of screen

EasyIP2023:
RTD theme's default is 800px as max width for the content, but we have tables with tons of columns, which need the full width of the view-port.

Comment from yocto project theme_overrides.css


recp:
@EasyIP2023 many thanks, the PR is merged 🚀

PR #412, opened 2024-03-31
separate SSE and SSE2

recp:
Separate **SSE** and **SSE2** where SSE2 may not be available e.g i686.

- [x] separate SSE and SSE2
- [x] make Fasth math work both on SSE and SSE2, make -0.0f run on SSE + fast math.
- [x] make some failing test pass when fast math is on 
- [x] provide `CGLM_FAST_MATH` to identify path math
- [x] tests: dont validate nan and inf on fast math

Most operations are SSE floating point but some SSE2 intrinsics are available for instance: 

- `_mm_castsi128_ps`
- `_mm_shuffle_epi32`

I tried to find some of them and fallback SSE FP domain if possible. 

Github action that only works on SSE would awesome actually to identify issues. 

**READY TO MERGE 🚀**

### DIFF
--- a/include/cglm/common.h
+++ b/include/cglm/common.h
@@ -45,6 +45,10 @@
 #  define CGLM_LIKELY(expr)   (expr)
 #endif
 
+#if defined(_M_FP_FAST) || defined(__FAST_MATH__)
+#  define CGLM_FAST_MATH
+#endif
+
 #define GLM_SHUFFLE4(z, y, x, w) (((z) << 6) | ((y) << 4) | ((x) << 2) | (w))
 #define GLM_SHUFFLE3(z, y, x)    (((z) << 4) | ((y) << 2) | (x))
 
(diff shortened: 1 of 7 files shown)


recp:
Big problem here:

Setting `-0.0f` is a big issue without SSE2 e.g. SSE + Fasth math which will ignore the sign probably. Also passing `0x80000000` directly to `float` seems not work.

recp:
This seems worked:

```C
#if defined(__SSE2__)
#  define GLMM_NEGZEROf ((int)0x80000000) /*  0x80000000 ---> -0.0f  */
#  define GLMM_POSZEROf ((int)0x00000000) /*  0x00000000 ---> +0.0f  */
#else
#  ifdef CGLM_FAST_MATH
     union { int i; float f; } static GLMM_NEGZEROf_TU = { .i = (int)0x80000000 };
#    define GLMM_NEGZEROf GLMM_NEGZEROf_TU.f
#    define GLMM_POSZEROf 0.0f
#  else
#    define GLMM_NEGZEROf -0.0f
#    define GLMM_POSZEROf  0.0f
#  endif
#endif
```

it defines `GLMM_NEGZEROf_TU` for each translation unit but better approaches can be considered if available

PR #416, opened 2024-05-11
added quat_slerp_longest

telephone001:
adds quat slerp that takes the longest path
I still need to figure out ideas for tests.
Let me know if I missed anything


telephone001:
https://www.youtube.com/watch?v=hOj2il2y-dM

recp:
Hi @telephone001,

Thanks for the implementation 🚀 

> Let me know if I missed anything

Tests would be awesome 😎 

> I still need to figure out ideas for tests.

A PR for test would be nice, later maybe when you have ideas ;)

Issue #344, opened 2023-08-25
Quaternion New Features 


recp:
https://developer.apple.com/documentation/accelerate/working_with_quaternions

We already have `glm_quat_rotatev()` ( which is similar to `simd_act()` I guess ), `glm_quat_rotate()` , `glm_quat_rotate_at()` ... but 

1. `glm_quat_spline()`
2. `glm_quat_bezier()`
3. `glm_quat_slerp_longest()`

would be nice to have.

telephone001:
thinking of implementing glm_quat_slerp_longest(versor q, versor r, float t, versor dest) by making r negative, comparing how big the angle is with q with r negative and regular r then slerping with the longer one.

telephone001:
Sorry I have been busy with some stuff and forgot to implement this. Hopefully someone else can implement this one.

recp:
@telephone001 thanks, I'm also busy for a while but can implement when available

telephone001:
I'm back and ready to finish these things. I just realized that it was extremely easy to implement glm_quat_slerp_longest. But I'm also ready to implement the other two.

recp:
Hi @telephone001,

Sorry for the delay,

Many thanks 🤗

PR #417, opened 2024-05-16
added a quaternion cubic bezier for cglm

telephone001:
Tried to make a bernstein polynomial for it but since each slerp depends on the angle, I don't think its gonna work.
So I used Decasteljau. Interested in hearing how I can improve this in the comments


telephone001:
just realized the bezier you requested isn't a cubic bezier so just gonna close this and do more research

recp:
Wow, sorry for the late response due busy days... 

I also need to do some research for `glm_quat_bezier()` and `glm_quat_spline()` as mentioned at https://github.com/recp/cglm/issues/344

PR #420, opened 2024-06-09
Fixed 'missing-prototypes' warnings for some vec2 funcs.

not-kaz:
Using gcc and missing-prototypes flag shows warnings for: glmc_vec2_fill, glmc_vec2_eq and glmc_vec2_eqv. 
Seems they were not added to call/vec2.h for some reason. This should fix the warnings.


recp:
@not-kaz the PR is merged, thanks for your contributions 🚀

PR #421, opened 2024-07-03
mat4: wasm simd128 for glm_mat4_inv

myfreeer:
The function `glm_mat4_inv_wasm` has been implemented, but not used in `glm_mat4_inv`. This commit adds a conditional macro to add the case of wasm and simd128 case for calling `glm_mat4_inv_wasm`.


recp:
@myfreeer the PR is merged, thanks for your contributions 🚀

PR #423, opened 2024-07-14
Reduce typo count.

waywardmonkeys:


recp:
@waywardmonkeys thanks for your contributions 🚀

PR #430, opened 2024-09-09
Bump CMake minimum version to 3.13

nitrix:
## What

This bumps the minimum CMake version from `3.8.2` to `3.13.0`.

## Goal

Hiding this warning message:

```
Not searching for unused variables given on the command line.
[cmake] CMake Warning (dev) at vendor/cglm/CMakeLists.txt:15 (option):
[cmake]   Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
[cmake]   --help-policy CMP0077" for policy details.  Use the cmake_policy command to
[cmake]   set the policy and suppress this warning.
[cmake] 
[cmake]   For compatibility with older versions of CMake, option is clearing the
[cmake]   normal variable 'CGLM_STATIC'.
[cmake] This warning is for project developers.  Use -Wno-dev to suppress it.
[cmake] 
```

## Why

Different versions of CMake comes with different policies enabled by default.

As of 3.13.0, CMake introduced the policy [CMP0077](https://cmake.org/cmake/help/latest/policy/CMP0077.html) which, because this project is on an older version, wants users vendoring `cglm` to specify whether they prefer the `OLD` or `NEW` behavior, resulting in either `set(CMAKE_POLICY_DEFAULT_CMP0077 X)` before `add_subdirectory(vendor/cglm)` or `cmake_policy(CMP0077 X)` for the whole project.

## Explanation

The `NEW` behavior (that version 3.13.0 introduced) is sane and desirable for everybody.

Prior, `option()` was copying identically named variables into the cache and then ignoring any future mutations unless re-configuring the whole project. Now, `option()` prefers to do nothing, not even caching, essentially honoring at all times what the early variable definition requested (that option() would've normally set).

This allows for users to properly configure CGML's options even though it's a subdirectory and to not worry about possible overrides or left over cache headaches.

e.g.

```
option(CGLM_STATIC "Static build" ON)
add_subdirectory(vendor/cglm)
target_add_library(example PRIVATE cglm)
```

## Backwards compatibility

* Users that explicitly want to keep the old behavior would've already needed the policy line and the version upgrade wont distrupt their choice. That's the whole point of configurable CMake policies.

* Everybody else on newer CMake versions that didn't pick a policy accidently rely on the new behavior and therefore not impacted by the new version because the default maintains that behavior.

## Notes

It seems 3.8.2 was in 2017 and 3.13.0 in 2018. Both are old enough that probably everyone and all the major distros are well beyond. Personally, I use MinGW a lot and am on 3.26.4, while officially the latest is 3.30.3.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.8.2)
+cmake_minimum_required(VERSION 3.13)
 project(cglm
   VERSION 0.9.5
   HOMEPAGE_URL https://github.com/recp/cglm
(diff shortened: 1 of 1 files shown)


recp:
Hi @nitrix, the PR is merged, thanks for your contributions 🚀

PR #442, opened 2025-01-20
Rename struct/aabb2d.h functions to match aabb2d.h, add tests

### LINKED ISSUE #429: aabb2d Struct API missing diag and sizev functions from Array API
The aabb2d Struct API has [glms_aabb2d_size](https://github.com/recp/cglm/blob/c12f318fabf14853174419f3892654612186c5c7/include/cglm/struct/aabb2d.h#L140C6-L140C19) which should probably be renamed to glms_aabb2d_diag and call [glm_aabb2d_diag](https://github.com/recp/cglm/blob/c12f318fabf14853174419f3892654612186c5c7/include/cglm/aabb2d.h#L169) from the Array API. It is also missing glms_aabb2d_sizev which should call [glm_aabb2d_sizev](https://github.com/recp/cglm/blob/c12f318fabf14853174419f3892654612186c5c7/include/cglm/aabb2d.h#L181).

duarm:
closes #429

### Problem

Missing _sizev and _diag from structs/aabb2d

### Solution

Rename _size to _diag, add missing _sizev, I've also added a test file for the aabb2d api

### Discussion

The struct API usually returns the struct instead of a dest parameter, aabb2d and box however, still have a dest parameter on some functions, as we do not have an aabb2d struct, just an array. Since the _sizev is a new function, I've opted for returning the vec2s instead of a dest parameter. Should I keep it this way or make it a dest parameter for consistency?

```c
void glms_aabb2d_sizev(vec2s aabb[2], vec2s *dest);
# or
vec2s glms_aabb2d_sizev(vec2s aabb[2]);
```


recp:
Hi @duarm,

No `dest` param in structure api, it looks good! 

Since it seems Draft, can we merge it?

duarm:
I was planning on adding more tests, but that can be done later, let's merge

PR #444, opened 2025-01-30
Fix numbering - opengl matrix uniform examples (docs)

wethu:
I was just reading the docs and noticed this, I think this should be numbered 2? Above is an alternative approach already numbered 1.


recp:
Hi @wethu,

The PR is merged, thanks 🚀

PR #452, opened 2025-03-11
Added Tau constant definitions to types.h

RolandMarchand:
This PR adds constant definitions for Tau (𝜏), which is 2 * pi, or 360 degrees in radians.

The following definitions are added for `float` and `double` types:
- `GLM_TAU` (tau)
- `GLM_TAU_2` (tau / 2)
- `GLM_TAU_4` (tau / 4)
- `GLM_1_TAU` (1 / tau)
- `GLM_2_TAU` (2 / tau)
- `GLM_SQRTTAU` (sqrt(tau))

**Note**: these constants were calculated with `bc`, a precision calculator language. This means that the constants do not include floating point imprecision errors.

### DIFF
--- a/include/cglm/types.h
+++ b/include/cglm/types.h
@@ -106,23 +106,35 @@ typedef vec3                    mat4x3[4];  /* [col (4), row (3)] */
 #define GLM_PI_4      0.785398163397448309615660845819875721  /* pi/4        */
 #define GLM_1_PI      0.318309886183790671537767526745028724  /* 1/pi        */
 #define GLM_2_PI      0.636619772367581343075535053490057448  /* 2/pi        */
+#define GLM_TAU       6.283185307179586476925286766559005768  /* tau         */
+#define GLM_TAU_2     GLM_PI                                  /* tau/2       */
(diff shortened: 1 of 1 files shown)


aaronfranke:
Awesome to see this PR merged! Note that "convert any angle to tau" is not really a meaningful statement. Angles may be in terms of radians, degrees, or turns. Tau is not a measurement, but is a constant. It is the constant used to convert between radians and turns (`turns * GLM_TAU = radians`).

> At most i'd agree with adding GLM_2PI (2*pi) since that's a commonly used constant,

The problem with that is Tau is a well-known constant, defined in dozens of languages and many tools/frameworks. It results in more readable code to use this name. Also, `2PI` can be confused with `2_PI` which means `2 / PI`. Perhaps you can memorize the meaning in one particular library, but can you go to another and know what it means without looking at the definition? Tau is not used for any other mathematical constant, so if you see Tau and are aware of it, the meaning is unambiguous. Merging this PR with the Tau name is the right move IMO (though some of the other values like `GLM_TAU_2` are debatable, likely helpful but not necessary).

PR #469, opened 2025-04-06
Feature/mat2 tables

EasyIP2023:


EasyIP2023:
@recp Re-arranged some functions and changed parameter names for API consistency.

Question with

```C
glms_mat2_(transpose)(mat2s m)
```
Do we need to break this up into

```C
mat2s
glms_mat2_(transpose_to)(mat2s m);

void
glms_mat2_(transpose)(mat2s m);
```

EasyIP2023:
@recp Okay good to know. Left that the same. Lastly wanted to update the docs format. So, that it's hopefully easier for new comers to understand the mathematics

1. Start with function definition, parameters, etc...
2. Given mathematical definition in table format
3. Provide an example in table format of a code snippet

Good? Or is it a bit too much?

Screenshot from 2025-04-06 10-56-11
Screenshot from 2025-04-06 10-57-22
Screenshot from 2025-04-06 11-00-03

recp:
@EasyIP2023 the docs looks good many thanks. But I'm not sure about naming about `src`.

`m|mat` -> `src`

for instance 

```C
glms_mat2_inv(mat2 src, mat2 dest)
```

we are inverting src okay but also inverting given matrix which is `m`. I would prefer m/mat and similar over src even it writes to `dest` param. 

Also just realized that `glms_mat2_inv()` should not have dest since dest is the returned value, see `glms_mat4_inv()` and `glms_mat3_inv()`

EasyIP2023:
@recp Okay cool cool. Will change functions params back when I find time and let you know.

EasyIP2023:
@recp Believe to be ready for another review.

PR #492, opened 2026-03-22
Fix "Call API pulls inline API"

### LINKED ISSUE #491: Call API pulls inline API
`#include "cglm/call.h` causes the inline API to be included, making the call API pointless. See the output of
`clang -I include/ -E - <<<"#include <cglm/call.h>" | grep '^glm_.*{'`

Including cglm/call.h should only include types and `glmc_` function declarations.

FunkyAss:
Fixes recp/cglm#491

### DIFF
--- a/docs/source/call.rst
+++ b/docs/source/call.rst
@@ -15,5 +15,6 @@ It would be duplicate documentation also it
 would be hard to sync documentation between inline and call version for me.
 
 By including **clgm/cglm.h** you include all inline versions. To get precompiled
-versions you need to include **cglm/call.h** header it also includes all
-call versions plus *clgm/cglm.h* (inline versions)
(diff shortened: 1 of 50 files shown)


recp:
Hi @FunkyAss,

thanks for your contributions,

- [ ] use relative includes *internally*: `cglm/` -> `../` to avoid collisions
- [ ] include/cglm/call/io.h must explicitly include <stdio.h>, common.h doesnt include io/FILE ( cglm.h was )
- [ ] we need to clarify the behavior change in docs/README which is inlines must be included separately than call

FunkyAss:
Roger that. I'll address the comments in the next (7-14) days

recp:
@FunkyAss the PR is merged, thanks for your contributions 🚀 sorry for the delay, it looks like in progress for a while...

